PR #26, opened 2014-10-10
preserve loaded file name for Save/Save As/Save A Copy file name proposals

varlog00:
See commit comment. It covers cases where input file folder is not valid anymore but the new file was read.
This is a second try (first after paths reworking).
 It is important for my usage case ( I've even nagged DiapDealer to get a meaningful file name from KindleImport plugin for this), but I think it is a general enhancement for Sigil. .

### DIFF
--- a/src/Sigil/MainUI/MainWindow.cpp
+++ b/src/Sigil/MainUI/MainWindow.cpp
@@ -3317,6 +3317,7 @@ void MainWindow::CreateNewBook()
     SetNewBook(new_book);
     new_book->SetModified(false);
     m_SaveACopyFilename = "";
+    m_CurrentFileName.clear();
     UpdateUiWithCurrentFile("");
 }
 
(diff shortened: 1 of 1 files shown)


varlog00:
Can I open it again (it is a test! just in case :))?

kevinhendricks:
sure

varlog00:
Apparently no :(. Closed by Higher Authority.

kevinhendricks:
It seems even I can not reopen it either.  Please grab your diff and see if it applies over master.  If so, please create a new pull request with it.

dougmassay:
From what I can tell, Merge Requests that are closed can never be reopened. Issues, yes. But Merge Requests are closed for good when closed/declined. Seems weird, but that's what I'm reading.

varlog00:
Just a hint for KindleImport :):

http://www.mobileread.com/forums/showthread.php?t=249533

PR #47, opened 2015-01-13
"Tools -> Clean Content" action

marek-g:
Hi!

I implemented new cleaning tool action for Sigil. It's useful for cleaning EPUBs converted from PDF files. The action is placed under Tools/Clean Content. It has 2 main options:

1. Remove page numbers.

You can provide sample page number string like "Page 12 of 34". It transforms it to "Page \d+ of \d+" regular expression and removes all matching paragraphs. This option is easier to use by people who don't know regular expression - they just copy/paste sample formatted page number.

2. Join paragraphs with split sentences.

Like "Join paragraphs" saved search but more intelligent. For example it automatically removes hyphenations.

Sigil has small but powerfull set of well thought features. So if you think my action doesn't fit to it - it's ok. It's very useful for me, so I thought I can share with you.

Best Regards,
Marek Gibek

### DIFF
--- a/src/Sigil/CMakeLists.txt
+++ b/src/Sigil/CMakeLists.txt
@@ -129,6 +129,8 @@ set( DIALOG_FILES
     Dialogs/About.h
     Dialogs/AddMetadata.cpp
     Dialogs/AddMetadata.h
+    Dialogs/CleanContent.h
+    Dialogs/CleanContent.cpp
     Dialogs/ClipboardHistorySelector.h
     Dialogs/ClipboardHistorySelector.cpp
     Dialogs/DeleteFiles.cpp
@@ -440,6 +442,7 @@ set( UI_FILES
     Form_Files/ReportsCharactersInHTMLFilesWidget.ui
     Form_Files/LinkStylesheets.ui 
     Form_Files/SearchEditor.ui
+    Form_Files/CleanContent.ui
     Form_Files/ClipEditor.ui
     Form_Files/IndexEditor.ui 
     Form_Files/SpellcheckEditor.ui 
--- a/src/Sigil/CMakeLists.txt
+++ b/src/Sigil/CMakeLists.txt
@@ -463,6 +463,8 @@ set( SOURCEUPDATE_FILES
     SourceUpdates/PerformCSSUpdates.h
     SourceUpdates/AnchorUpdates.cpp
     SourceUpdates/AnchorUpdates.h
+    SourceUpdates/CleanContentUpdates.cpp
+    SourceUpdates/CleanContentUpdates.h
     SourceUpdates/LinkUpdates.cpp
     SourceUpdates/LinkUpdates.h
     SourceUpdates/WordUpdates.cpp
--- a/src/Sigil/Dialogs/About.cpp
+++ b/src/Sigil/Dialogs/About.cpp
@@ -58,6 +58,7 @@ About::About(QWidget *parent)
                       "<li>Charles King</li>" +
                       "<li>Daniel Pavel</li>" +
                       "<li>Grzegorz Wolszczak</li>" +
+                      "<li>Marek Gibek</li>" +
                       "</ul>" +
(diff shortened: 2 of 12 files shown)


kevinhendricks:
Hi,
Cleaning leftovers from pdf conversions is probably better suited for a Sigil plugin as it fits a very specific need and not a general use case.  Please consider creating a python2 or python3 based Sigil plugin instead.  Ask if you need more information on how to create plugins.

marek-g:
Hi Kevin!

Probably you are right. I thought about it at the beginning but C++
code works better for me:

1. It's no work for me to create portable Sigil (on Windows platform)
- it would be a problem with python installation (also my father uses
it and he hates to install additional components on his PC and loosing
portable aspect of the app or making the app triple as big as the
previous version ;)).

2. C++ code works faster - it's a beauty to use it on 8 threads in
instant on whole ebook! I do not expect the same performance with
python.

3. It was an opportunity for me to familiarize myself with Sigil
source code. My next step is to extend metadata editor - allow WYSIWYG
HTML editing and fetching meta data from web page (because calibre
doesn't write metadata back to epub file - I consider extending
calibre instead - I haven't decided yet). Extending UI isn't probably
possible with python plugin yet so I didn't see any point to invest in
learning it...

I see your point, Kevin. And I agree with it, so I haven't expected
too much :) However, I don't plan to rewrite it in Python (especially
that I plan to extend Sigil further for my needs - it's the prefect
base for me). So, please close this pull request and the code will
stay only in my branch :)))

Thanks for quick answer!
Marek

kevinhendricks:
Hi Marek,

FWIW, Linux and Mac OS X comes with system python installed so nothing to add or move or that add size to anything.  Most Windows serious about ebook development have either python2 already on their systems and most of the ebook conversion tools are written in python (ie. KindleUnpack, calibre, etc). The plugin code can actually be run from the system based python interpreter and as such does not hurt portability or size. That is one of the key reasons we decided to go with python as the plugin language for Sigil.   If you want help developing a python based plugin for Sigil, let me know and I would be happy to point you to the right resources.

Also please be aware that Sigil is undergoing some significant changes to prepare it for epub 3 support over the coming year and Metadata will of course be one of the areas impacted.  If you are looking for a project and would like to become more involved with Sigil development, please let us know.  

Rewriting the Metadata code to support the new epub3 metadata format would certainly be a helpful project if you are at all interested.  

Additional projects that would really help us out would be:

1. pulling flightcrew out of the Sigil build and replace it with a standalone plugin version

2. replace Tidy with a better solution probably based on google's gumbo-parser (but that has not been finally determined yet) that is html5 aware and that will be robust/self-repairing for poor html code

3. update the opf parsing and generation code to support the new features in the opf 3 spec.

Thanks,

Kevin

marek-g:
Hi Kevin!

Thanks for all. I would love to help extend Sigil as it is the best
epub editing tool I know. It is better than commercial products I've
tried. It looks really solid. But working full time and having my own
million ideas waiting to be coded after work - sadly I have no time
for this :'( And Sigil is pretty complete for me and very very useful
without any changes.

So I'm limiting myself to features I'm really missing.

Funny, but I'm not waiting for epub3 support at all. I'm using e-Ink
reader for reading - I don't expect it to play videos :) I don't think
that books and multimedia matches well. I can watch multimedia
presentations on a tablet with HTML5 browser. For interactivity I
prefer specialized players like interactive fiction. Maybe there is
something more interesting in epub3 but for me epub2 has more than
enough to support all books I want to read.

Anyway, I may be adding small fixes to things that are annoying me, for example:
- series / series number missing in meta editor (probably a low
hanging fruit ;))
- sometimes it is hard to control bold/italic styles using WYSIWYG
editor (don't know yet if it's a problem with Qt web control or Sigil
or with too many ways to express the same in HTML)
etc.
but I don't see many of them... Great work guys!

Best Regards,
Marek

marek-g:
Hm... I may be missing one thing from EPUB3 only - it's MathML :)

PR #48, opened 2015-01-16
Changes for LMDE 64bit, Qt 5.1.1

tasptz:
Just wanted to let you know that Sigil 0.8.2 compiles fine on Linux Mint Debian 64bit (Qt 5.1.1) with two minor changes.

### DIFF
--- a/src/Sigil/Form_Files/PluginRunner.ui
+++ b/src/Sigil/Form_Files/PluginRunner.ui
@@ -86,7 +86,7 @@ p, li { white-space: pre-wrap; }
        <property name="textInteractionFlags">
         <set>Qt::TextSelectableByKeyboard|Qt::TextSelectableByMouse</set>
        </property>
-       <property name="placeholderText">
+       <property name="placeholderText" stdset="0">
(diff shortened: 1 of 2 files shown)


kevinhendricks:
Hi,
Sigil 0.8.2  was built against Qt 5.3 and master is built with Qt 5.4.0 due to the large number of bug fixes  since the earlier versions of Qt 5.  Qt 5.4 is available as a binary installer for Linux direct from the Qt website.  Sigil master will build out of the box with CMake 3 and Linux Mint and Qt 5.4.0 (this is one of the VM builds I build to check Sigil).

In addition, master has workarounds to avoid some serious memory corruption issues in Qt.  So I strongly recommend using at least Qt5.4.0 and Sigil's current master if at all possible. Neither of your changes is needed by any of the newer Qt releases since 5.2 I believe.

Kevin

kevinhendricks:
Closing this as Sigil is now using Qt5.4.0 as part of its main release which needs none of these changes and brings with it a large number of Qt bug fixes.

PR #51, opened 2015-01-24
Cleaned up source code.

jwvdveen:
Cleaned up source code.

### DIFF
--- a/src/Sigil/MainUI/TableOfContents.cpp
+++ b/src/Sigil/MainUI/TableOfContents.cpp
@@ -19,19 +19,20 @@
 **
 *************************************************************************/
 
-#include <QFileInfo>
-#include <QTimer>
-#include <QTreeView>
-#include <QContextMenuEvent>
-#include <QAction>
-#include <QMenu>
+#include <QtCore/QFileInfo>
+#include <QtCore/QTimer>
+#include <QtWidgets/QTreeView>
+#include <QtWidgets/QVBoxLayout>
+#include <QtGui/QContextMenuEvent>
+#include <QtWidgets/QAction>
+#include <QtWidgets/QMenu>
 
-#include "../BookManipulation/FolderKeeper.h"
-#include "../MainUI/TableOfContents.h"
-#include "../Misc/Utility.h"
-#include "../ResourceObjects/NCXResource.h"
-#include "../sigil_constants.h"
-#include "../sigil_exception.h"
+#include "BookManipulation/FolderKeeper.h"
+#include "MainUI/TableOfContents.h"
+#include "Misc/Utility.h"
+#include "ResourceObjects/NCXResource.h"
+#include "sigil_constants.h"
+#include "sigil_exception.h"
 
 static const int COLUMN_INDENTATION = 15;
 static const int REFRESH_DELAY = 1000;
@@ -40,17 +41,31 @@ TableOfContents::TableOfContents(QWidget *parent)
     :
     QDockWidget(tr("Table Of Contents"), parent),
     m_Book(NULL),
-    m_TreeView(*new QTreeView(this)),
+    m_MainWidget(*new QWidget(this)),
+    m_Layout(*new QVBoxLayout(&m_MainWidget)),
+    m_TreeView(*new QTreeView(&m_MainWidget)),
     m_RefreshTimer(*new QTimer(this)),
     m_NCXModel(*new NCXModel(this))
 {
-    setWidget(&m_TreeView);
+    m_Layout.setContentsMargins(0, 0, 0, 0);
+#ifdef Q_OS_MAC
+    m_Layout.setSpacing(4);
+#endif
+    m_Layout.addWidget(&m_TreeView);
+    m_MainWidget.setLayout(&m_Layout);
+    setWidget(&m_MainWidget);
     m_RefreshTimer.setInterval(REFRESH_DELAY);
     m_RefreshTimer.setSingleShot(true);
     SetupTreeView();
     connect(&m_TreeView, SIGNAL(clicked(const QModelIndex &)), this, SLOT(ItemClickedHandler(const QModelIndex &)));
     connect(&m_RefreshTimer, SIGNAL(timeout()), this, SLOT(Refresh()));
     connect(&m_NCXModel, SIGNAL(RefreshDone()), &m_TreeView, SLOT(expandAll()));
+}
+
+void TableOfContents::showEvent(QShowEvent *event)
+{
+    QDockWidget::showEvent(event);
+    raise();
 }
 
 void TableOfContents::SetBook(QSharedPointer<Book> book)
@@ -68,8 +83,10 @@ void TableOfContents::Refresh()
 
 void TableOfContents::StartRefreshDelay()
 {
-    // Repeatedly calling start() will re-start the timer and that's exactly what we want.
-    // We want the timer to fire REFRESH_DELAY miliseconds after the user has stopped typing up the NCX.
+    // Repeatedly calling start() will re-start the timer
+    // and that's exactly what we want.
+    // We want the timer to fire REFRESH_DELAY miliseconds
+    // after the user has stopped typing up the NCX.
     m_RefreshTimer.start();
 }
 
@@ -80,10 +97,11 @@ void TableOfContents::RenumberTOCContents()
 
 void TableOfContents::ItemClickedHandler(const QModelIndex &index)
 {
-    QUrl url = m_NCXModel.GetUrlForIndex(index);
+    QUrl url         = m_NCXModel.GetUrlForIndex(index);
     QString filename = QFileInfo(url.path()).fileName();
 
     int line = -1;
+
     // If no id, go to the top of the page
     if (url.fragment().isEmpty()) {
         line = 1;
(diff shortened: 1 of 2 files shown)


user-none:
This is not correct at all. The includes should not be changed to relative paths. The build system has the paths for them as part of it and they are _not_ supposed to be referenced via their location on the file system. This is specifically so they can be relocated if necessary. The build system would be updated to include new or modified to set the appropriate path for the compiler to look. Simply put, have the path to a header referenced to a location on the filesystem is not okay or correct in any way.

Sigil is multi-platoform changes like the block with  `m_TreeView.setEditTriggers` may not seem necessary on your platform but might be very much required on another. This also goes into changes like removing the Mac OS  X spacing define.

Some parts are fine like removing the `GenerateTocRequest` signal that's never used.

That said in this state it cannot be merged.

jwvdveen:
Ah, I see. Although I did nog changed the includes at first, after I installed Qt5.4 instead of Qt5.3.2, QtCreator could not verify the locations and so I lost things as "Find usages". Although it did compile.
Soms of You have experience with that?

Verzonden met mijn Windows Phone
________________________________
Van: John Schember<mailto:notifications@github.com>
Verzonden: ‎25-‎1-‎2015 00:41
Aan: user-none/Sigil<mailto:Sigil@noreply.github.com>
CC: jwvdveen<mailto:jwvdveen@outlook.com>
Onderwerp: Re: [Sigil] Cleaned up source code. (#51)

This is not correct at all. The includes should not be changed to relative paths. The build system has the paths for them as part of it and they are _not_ supposed to be referenced via their location on the file system. This is specifically so they can be relocated if necessary. The build system would be updated to include new or modified to set the appropriate path for the compiler to look. Simply put, have the path to a header referenced to a location on the filesystem is not okay or correct in any way.

Sigil is multi-platoform changes like the block with  `m_TreeView.setEditTriggers` may not seem necessary on your platform but might be very much required on another. This also goes into changes like removing the Mac OS  X spacing define.

Some parts are fine like removing the `GenerateTocRequest` signal that's never used.

That said in this state it cannot be merged.


---

jwvdveen:
By the way, the reason I removed the Mac OS X spacing define was because I added the m_TreeView directly to the Table of contents DockWidget. If I recall it correct, the spacing define was about the layout which I had removed.

Also, doe different OSses have different default values? If Qt had a default value "false", could this on Mac OS X be "true"? I never gave this much thought before, since Linux and Windows always gave me the same results so far.

Verzonden met mijn Windows Phone
________________________________
Van: John Schember<mailto:notifications@github.com>
Verzonden: ‎25-‎1-‎2015 00:41
Aan: user-none/Sigil<mailto:Sigil@noreply.github.com>
CC: jwvdveen<mailto:jwvdveen@outlook.com>
Onderwerp: Re: [Sigil] Cleaned up source code. (#51)

This is not correct at all. The includes should not be changed to relative paths. The build system has the paths for them as part of it and they are _not_ supposed to be referenced via their location on the file system. This is specifically so they can be relocated if necessary. The build system would be updated to include new or modified to set the appropriate path for the compiler to look. Simply put, have the path to a header referenced to a location on the filesystem is not okay or correct in any way.

Sigil is multi-platoform changes like the block with  `m_TreeView.setEditTriggers` may not seem necessary on your platform but might be very much required on another. This also goes into changes like removing the Mac OS  X spacing define.

Some parts are fine like removing the `GenerateTocRequest` signal that's never used.

That said in this state it cannot be merged.


---

user-none:
> Also, doe different OSses have different default values?

Yes.

> If you're using Qt Creator (which I do not) then Qt Creator will set the same default value for elements created using it.

user-none:
Also the layout is necessary on OS X to get the spacing and margins proper.

user-none:
OS X is it's own special animal and a lot of work is done (and has to be done) for things to appear proper on it. It's gotten better of the years but without testing I can't say if remove the layout would break the way it looks on OS X. Also realize that primary development happens on OS X so while something like that layout may seem strange or arbitrary it's probably because that's the only way to get Qt to display properly on OS X.

jwvdveen:
Ah, I see. Sorry for the trouble. I will be more carefull and recommit these files again next week.

And I will try to find the solution of Qt5.4 not finding the relative include paths (which Qt5.3.2 did).

Thanks for clearing things up. Now I will go to sleep since it's in the middle of the night here.

Verzonden met mijn Windows Phone
________________________________
Van: John Schember<mailto:notifications@github.com>
Verzonden: ‎25-‎1-‎2015 01:41
Aan: user-none/Sigil<mailto:Sigil@noreply.github.com>
CC: jwvdveen<mailto:jwvdveen@outlook.com>
Onderwerp: Re: [Sigil] Cleaned up source code. (#51)

OS X is it's own special animal and a lot of work is done (and has to be done) for things to appear proper on it. It's gotten better of the years but without testing I can't say if remove the layout would break the way it looks on OS X. Also realize that primary development happens on OS X so while something like that layout may seem strange or arbitrary it's probably because that's the only way to get Qt to display properly on OS X.

---

user-none:
You should look at the docs folder it has a number of ways to build Sigil. I just built it against Qt 5.4.0 on OS X via Unix Make files and no issues... So this could be a bug with Qt Creator or it could be something with how it interacts with CMake.

Can you provide more info about your platform and setup? Do you have multiple Qt versions installed for example? Are you calling CMake yourself or relying on QtCreator. Are you working off of a previous build before updating to Qt 5.4.0?

If I remember correctly when ever you change Qt versions you need to blow out the build dir and rebuild from scratch. You also need to reconfigure.

Also, I do know I need to explicitly tell CMake were to look for Qt on my system, `-DCMAKE_PREFIX_PATH=/usr/local/opt/qt5/lib/cmake/`

jwvdveen:
Thanks. The problem is not Qt, but seems to be QtCreator 3.3.0. So I 
reverted to Qt5.3.2 with QtCreator 3.2.1 and everything works fine 
again. Only issue is a comment about  QtWebChannel but this does not 
prevent the building of Sigil.

About the layout of Mac OS X, wouldn't something like QMacStyle or 
app.setStyle(QStyleFactory::create("macintosh")) handle those things. Or 
even a custom style? It is just a thought however.

I am using Debian Jessie 64bit with Qt5.3.2 which I downloaded from the 
website of Qt and cmake3.0.2.


Op 25-01-15 om 01:56 schreef John Schember:

jwvdveen:
By the way, I just re-added the table of contents source code. I hope it is correct now.

Verzonden met mijn Windows Phone
________________________________
Van: Jan-Willem<mailto:jwvdveen@outlook.com>
Verzonden: ‎25-‎1-‎2015 12:09
Aan: user-none/Sigil<mailto:reply@reply.github.com>
Onderwerp: Re: [Sigil] Cleaned up source code. (#51)

Thanks. The problem is not Qt, but seems to be QtCreator 3.3.0. So I
reverted to Qt5.3.2 with QtCreator 3.2.1 and everything works fine
again. Only issue is a comment about  QtWebChannel but this does not
prevent the building of Sigil.

About the layout of Mac OS X, wouldn't something like QMacStyle or
app.setStyle(QStyleFactory::create("macintosh")) handle those things. Or
even a custom style? It is just a thought however.

I am using Debian Jessie 64bit with Qt5.3.2 which I downloaded from the
website of Qt and cmake3.0.2.


Op 25-01-15 om 01:56 schreef John Schember:

user-none:
> About the layout of Mac OS X, wouldn't something like QMacStyle or
> app.setStyle(QStyleFactory::create("macintosh")) handle those things. Or
> even a custom style? It is just a thought however.

This was integrated into Qt proper so you don't have to to this any more at all with Qt5. It's automatic and not something that you need to enable (or can disable). I believe you're looking at the Qt4 docs (understandable as they are typically first in Google since Qt 4 was around for so long). You really need to be sure you're looking at the Qt5 docs as thing like the above are no longer relevant.

This is where OS X get's hard for the Qt people. You have a platform with a very different looks, feel, and interaction perspective than Windows or Linux. They're trying to mesh that into Qt so a user won't know that they're using a Qt app. They've come a long way but, a lot of times little things show though and the application developer (us) need to make special code changes to ensure the same behavior is maintained across platforms.

> I am using Debian Jessie 64bit with Qt5.3.2 which I downloaded from the
> website of Qt and cmake3.0.2.

Be sure (which I think you already did) to upgrade to Qt 5.4.0 or later (I don't think there is an or later but you get the idea). Qt 5.4 fixes a lot of bugs that were reported because while Sigil has it's own bugs a lot of bugs are really upstream issues wither Qt or WebKit (which Qt uses). Each release brings fixes. Sigil is a very use the latest project because we want to take advantage of things like Qt bug fixes.

The only exception is cmake. I'm trying to standardize on the 3.0.2 docs for the time being to ensure a consistent build base. That way anyone running cmake 3.0.2 or higher won't have an issue. That said if a newer version of cmake has some amazing feature I want to use then that will be a hard requirement.

In regard to cmake I don't think I use any 3.0.2 specific feature and the last 2.8 version will still work to build Sigil. In this case it's purely that I'm using the 3.0.2 docs so I can 100% say that if you use 3.0.2 or higher that will work. It gets difficult to say what version you need (other than saying the latest) if you're always using the latest docs because you don't know if your using a feature that is in an older version. I've noticed that with Qt at least they mark when new features were added and in what version right in the developer docs. I haven't seen that with cmake so I'm just using the 3.0.2 docs and saying use that because I know it will work.

Another thing I want to point out since you're using Debian. On Linux the default is to use the system provided libraries and fall back to the bundled version if available. When you look at the code you see things like boost, hunspell and so forth. This is something that distros want for their packaging and we rely on them to test for version incompatibilities since they want to have it broken up.

For Sigil development alway use the cmake build flag `-DFORCE_BUNDLED_COPIES=1`. This will ensure the bundled version of the libs are used when you build. This way we know if there is an issue it's not with a conflict with a system library or a bug introduced in a older or newer version of a system library.

user-none:
The new changes look good. Thanks for the help!

PR #56, opened 2015-02-12
Minor fix & Update link

kant:


user-none:
FYI, the old link while redirects to the new one. So, it still works. But you are correct the location did change and we should list the proper location.

PR #57, opened 2015-02-15
linux: install real binary in /lib/sigil/ instead of /share/sigil/

mapreri:
This is to comply with the [FHS](http://www.pathname.com/fhs/)

/cc @dougmassay 

This "fix" commit 1ebe42c330de5381d67f3c678eac62a23fb6c656

Thanks.


dougmassay:
The Sigil binary has always gone in {INSTALL_PREFIX}/bin. Since it's inception.

I am willing to put the real binary back there (alongside the sigil launch script) for the normal "make install" target and adjust the launch script accordingly. But that's about it.

Ultimately, though, it's not really my call.

user-none:
@mapreri FHS isn't complied by the vast majority of Linux distros out there. It is a standard that basically no one follows.

@dougmassay The Sigil binary should be in {INSTALL_PREFIX}/bin. You are correct about that.

As for the wrapper script it is specifically for instances where components are not part of the base installation. A good example was the switch to Qt5 where the vast majority of distors hadn't packaged it and it defaults installs to /opt when using the Qt-Project provided installer.

There are a few solutions I see for this issue.

1. Wrapper script is sigil in bin and sigil becomes sigil-bin. This happens with a lot commercial Linux software so it's not an unprecedented solution.
2. Sigil binary would go into bin. The wrapper script is put into /usr/share/sigil and any distro or package that needs it can use it as a base as necessary. This would cause an issue with current targets in Sigil's auto packaging for Linux distros. Which leads us to the next resolution.
3. The merge request gets merged but as noted this breaks existing functionality. This would need to be addressed before this could be merged and existing functionality is preserved.

One thing I've been thinking about is having a global Sigil "share" environment variable instead of separate ones for each individual thing. There are a lot that the wrapper script doesn't even cover and the total number is starting to get out of hand anyway. This way the wrapper script could point to one location that includes things like default dictionaries, clips examples, plugin framework...

user-none:
I should mention that I do agree with putting the binary in lib rather than share, if we are going to use the wrapper script by default on Linux installations. Packagers can tweak the wrapper script from there.

Optimally my vote is for number 2 above.

dougmassay:
I'm OK with any of the 3 options with regard to a typical Linux "make install."

Building the standalone binary package, however (which is absolutely non-standard to begin with), becomes overly complicated when the sigil binary and the various cherry-picked Qt plugin directories are not in the same directory. I would need to put a qt.conf file in the binary's new directory to tell the cherry-picked Qt libs where the cherry-picked Qt plugins are located. Not impossible, but then any other Qt program's binary that might be installed in /bin or /lib could also attempt to use that qt.conf file with potentially adverse results.

If the binary must be moved to /lib or /bin when building the standalone binary sigil package, also ... then I'll probably have to abandon the binary package target altogether. Which is fine if that needs to happen--I can easily go back to patching the CMakeLists.txt file to create standalone binary packages.

dougmassay:
@user-none If you're OK with using the launch script (sigil in INSTALL_PREFIX/bin/) by default on Linux installations, and putting the actual binary (sigil-bin) in INSTALL_PREFIX/lib/sigil/, then feel free to merge @mapreri 's PR. With a couple of quick, easy tweaks, I can ensure that plugin functionality (in both the default Linux installation and the binary package creation) is not compromised.

user-none:
I've merged this.

There are a few changes I made that I had talked about already and I've put into the docs file. but to reiterate so yo two can make any necessary changes. All of the separate env variables have been combined into one "SIGIL_EXTRA_ROOT". This is by default /usr/share/sigil. This is set in the launcher script and it uses CMAKE_INSTALL_PREFIX instead of just /usr. In case the installation is in /opt or /usr/local.

This does mean that on Linux the installation structure is rigid:

* bin/wrapper
* lib/sigil/sigil
* share/sigil

@dougmassay you'll want to ensure that make install on Linux installs the files into those locations. At the very least wrapper should be named sigil. lib/sigil/sigil could be that or sigil-bin. Doesn't really matter as long as lib/sigil/sigil isn't called directly because the wrapper sets up the environment.

The Qt Doc state that getting the application dir path using "applicationDirPath()" has bugs and could lead to the wrong dir. It could use bin/ or it could use lib/sigil. If it uses bin/ the default location paths will work. If it uses lib/sigi/ then they won't work so the wrapper script becomes very important on Linux since it sets the correct path always.

Make any additional changes you need and if necessary a lot of the places getenv was used we actually searched multiple paths so we could extend that to include ../../share if needed. There are few places that don't search but assume like in the plugin launcher location code but that can be easily changed. Let me know what's needed at this point if anything that goes for @mapreri too. Like I said I can change the search paths so it will work even if the binary is run directly.

user-none:
I went ahead and made the changes to accommodate looking if sigil is installed in base/share/sigil properly if the env wrapper isn't used. If base/lib/sigil is run it should still work... At least it will find the helper files needed for Sigil to run.

dougmassay:
Looks good. I do believe, though, that binary package launch script (sigil-sh_pkg) will still need to have the QT_PLUGINS_PATH env var set to match the new SIGIL_EXTRA_ROOT. Qt won't look in Sigil's 'helper' directory (install_base/share/sigil) for its builtin plugins by default anymore now that the Sigil binary isn't there. Or should I simply consider putting all the Qt5/icu libraries/plugins in the 'install_base/lib/sigil/' directory when whenever building the standalone binary package?

One extra thought  ... 

If the Sigil binary is going to be installed to "install_base/lib/sigil/sigil" by default, does it make sense to leave the hard-coded path of:

launcher_root += "/../share/" + QCoreApplication::applicationName().toLower() + "/plugin_launchers/";

in PluginDB.cpp and EmbeddedPython.cpp if 1) it's not going to be utilized when the launch script that sets SIGIL_EXTRA_ROOT is default, and 2) it's no longer a valid path in relation to where the sigil binary will be now, anyway. Could that hard-coded path not be changed to:

launcher_root += "/../../share/" + QCoreApplication::applicationName().toLower() + "/plugin_launchers/";

Then environment variables wouldn't really even be necessary to locate the 'plugin_launcher' and 'python3lib' directories (except in the event that someone manually relocates things).

Just a thought. I see you're using SIGIL_EXTRA_ROOT to find some other directories on Linux, so it's no real skin off my back if you keep it the way it is. It's just that that hard-coded path in those two *.cpp files seems a bit vestigial and rather pointless as things stand.

user-none:
@dougmassay 

> launcher_root += "/../../share/" + QCoreApplication::applicationName().toLower() + "/plugin_launchers/";

You're looking at an older commit. Everything was changed to use a list of paths and the first one it finds it uses. Also QCoreApplication::applicationName().toLower()  was changed to "sigil" instead of driving it because if they're running sigil-bin then that wouldn't work right. We're requiring a set dir structure now so hardcoding sigil instead of deriving should be fine.

Next up is QT_PLUGINS_PATH. I don't think I changed that one and I don't see it as ever having been set...

> Or should I simply consider putting all the Qt5/icu libraries/plugins in the 'install_base/lib/sigil/' directory when whenever building the standalone binary package?

I think that's probably the best idea @mapreri  does that work for you?

Also, I think that the make install target should mimic the package install locations.

dougmassay:
@user-none
 
>Next up is QT_PLUGINS_PATH. I don't think I changed that one and I don't see it as ever having been set...
It was never necessary, as the Sigil binary was in the same directory as the Qt libs/plugins before (only concerns the binary package build, btw). It won't become necessary either, if moving the Qt stuff to /lib/sigil/ along with the binary works as it should.

>I think that's probably the best idea @mapreri does that work for you?
This makes sense to me too. I'll experiment and make sure there's no problems.

>Also, I think that the make install target should mimic the package install locations.
With the exception of the Qt stuff the package install includes, this should indeed be the case when all is said and done.

mapreri:
@user-none thanks. that's completely fine by me.
@dougmassay thank you too!

Everything sounds much better now. I'll let you know whenever I have something else to complain.

PR #61, opened 2015-02-28
Tidy Linux build changes

dougmassay:
Tidied up a few things with the Linux binary relocation and package building and the SIGIL_EXTRA_ROOT env variable changes.

launch script = install_base/bin/sigil
sigil binary =  install_base/lib/sigil/sigil

QT libs to install_base/lib/sigil/ when binary package is built.

Launch scripts adjusted and tidied.

Fixed a few Linux compilation issues related to the addition of sigil_constants.cpp.


user-none:
committed. I removed the extern from the sigil_constants.cpp file because it shouldn't be needed there.

PR #68, opened 2015-04-22
Capitalize pcre directory name in the source code

kytvi2p:
Without this change (or without changing the capitalization of the `pcre` directory in the source), compiling will fail on case-sensitive filesystems.

Please consider applying this patch (or something similar) to make it easier for Linux users to build. Th


user-none:
From the looks of this pull it would be easier to just rename the die to be
all uppercase. That is unless git is case insensitive too.

kytvi2p:
Yes, renaming the directory should work as well.

PR #73, opened 2015-05-09
Introduce SHARE_INSTALL_DIR to install arch-independent data

heirecka:
This is helpful on a multiarch layout where the prefix is
/usr/${host} but arch-independent files should still be installed
into /usr/share.


user-none:
SHARE_INSTALL_DIR should be set in the Linux specific section because it does not apply to any other platform.

heirecka:
Sorry. As a packager of a Linux distro I sometimes seem to forget that there are other platforms. I updated the pull request.

user-none:
I'm still only seeing one commit. Did you forget to push it?

heirecka:
I amended the commit and force pushed. I never had to update a pull request on github before, next time time I'll add an extra commit.

user-none:
There were some issues I saw with the pull request. I pushed up a variation that should resolve your issue.

PR #78, opened 2015-06-01
Update sigil_zh_CN.ts

Inori-Lover:
更新中文翻译


user-none:
Translations are handled on https://www.transifex.com . They are never merged directly but pulled periodically (before release) from the translation project.

PR #118, opened 2015-09-23
CharToEntity what PerformHTMLUpdates returns

dougmassay:


kevinhendricks:
Hi Doug,

Thanks!

- I have pushed a change to the README.md to references your new building docs for Linux
- I have pushed a change to flightcrew for the plugin to work on python3 (required an o for octal constants for os.chmod)

I am sure we are going to want to make a few changes to flightcrew plugin code to detect 64 bit vs 32 bit on Windows and Linux so that the right versions are run.  We should end up with 3 FlightCrew plugins:

FlightCrew_v0.9.0osx.zip
FlightCrew_v0.9.0win.zip
FlightCrew_v0.9.0unx.zip

I have the Mac one ready to go but would like the exact same plugin.py code to be in each so I will wait for any needed changes before making the FlightCrew build.

Thanks,

Kevin

PR #121, opened 2015-09-24
Add install parameter to command to install cmake

Dave-ts:
Sorry, I should have done both the .html and the .md file at the same times rather than generating a second pull request. First time making a change and attempting to contribute to a project that wasn't mine in any way other than sending emails.


kevinhendricks:
Thanks for helping us out.

Issue #134, opened 2015-10-13
allow configuring a directory from which load hunspell dictionaries


mapreri:
In Debian we have a role to do not embed anything that is in another package.  This is true for hunspell dictionaries too.
Till the current stable there was a env variable usable to tell sigil where to look for hunspell dictionaries, and I added https://anonscm.debian.org/cgit/collab-maint/sigil.git/commit/?id=eef2181af4fe87fdd11f0c87a8cddddd092f250e to achive it nicely.

For now (still in a experimental branch) the only thing I could do is to add a symlink to that directory (which is imho less nice).

Would it be possible to re-add that "feature", or any other mean to specify a directory from where to look for dictionaries?  I'm ok also with build-time flags, and willing to write a patch for that if you tell me what you'd prefer.

kevinhendricks:
Well, you will note that Sigil has its own highly modified version of google's gumbo parser (modified to support xhtml parsing, more tags recognized, and editing), and our own highly modified version of the BeautifulSoup4 python package (modified to fix bugs in builder/_lxml.py, add in real prettyprint, and rework the entire package to work under both python 2, and python 3, at the same time).  If you try to replace either of these with official versions Sigil will fail to work.

And no, neither project will want to upstream our changes as they are specific to mainly ebooks.

I will ask DiapDealer (he does all of the Linux dev work for Linux).  I know he pulled a bunch of build related stuff out to greatly simplify things.  To many linux distributions use too many different conventions, to make things work, disks are cheap, and embedding you own prevents breakage from updating shared libraies and things.  Too bad Linux does not have a true, sandboxed app container and some standard conventions to work with them.

We'll get back to you on the Linux front.

Thanks,

KevinH

mapreri:
On Tue, Oct 13, 2015 at 04:00:02PM -0700, Kevin Hendricks wrote:
> Well, you will note that Sigil has its own highly modified version of
> google's gumbo parser (modified to support xhtml parsing, more tags
> recognized, and editing), and our own highly modified version of the
> BeautifulSoup4 python package (modified to fix bugs in
> builder/_lxml.py, add in real prettyprint, and rework the entire
> package to work under both python 2, and python 3, at the same time).
> If you try to replace either of these with official versions Sigil
> will fail to work.

yeah, I'm very well aware of this, sadly.  But I'm already happy enough
all the stuff in 3rdparty can be replaced by system/vendor provided
ones.
Please continue to work like this, because the tred you follow was:
[...]

PR #139, opened 2015-10-18
linux: use the system hunspell/hyphen dictionaries.

eli-schwartz:
When USE_SYSTEM_LIBS is active, configure sigil to use the standard system directories for hunspell and hyphen dictionaries, instead of installing its own dictionaries.
Also, move the vendored dictionaries into 3rdparty.

This is a better fix for #134.


dougmassay:
Hey Eli,

"Better Fix" is subjective ;)

I'm fine with moving the bundled dictionaries to "3rdparty." But I'd rather leave the alternate system hunspell dictionary locations as run-time environment variable overrides. Namely because I'm not crazy about including hard-coded paths to system hunspell stuff in the source tree. And partly because I don't want USE_SYSTEM_LIBS to _automatically_ mean that the system dictionaries have to be used instead of the bundled ones. The env var approach provides a little more flexibility in that regard. 

Nor am I crazy about making wrapper.py a file that needs to be configured first. It's important for the plugin framework to remain a standalone (and fully functional) unit which can be easily released separately if need be.

If you'd like to take a stab at reworking this system hyphen directory location thing utilizing the current SIGIL_DICTIONARIES env var approach--which should eliminate the need to make very many changes at all to wrapper.py,--I'll be happy to take a look.

Thanks,
Doug

eli-schwartz:
Okay, "better fix" is subjective. :D But I do think it is better to be able to handle via a configuration rather than via patching plus deleting files from the installation tree.

There is already one env var override (SIGIL_EXTRA_ROOT), so I don't see why we need another one. Configuring the plugin wrapper also means we don't even have to worry about "# If someone didn't launch Sigil with its launch script, this may save the day (as long as the user didn't override SHARE_INSTALL_PREFIX at compile time)."
But I guess I hear the need to keep it separately usable.


I wanted to handle it via USE_SYSTEM_LIBS because the whole purpose of that option is for people who want to debundle what can be debundled :p but I guess I could change it to use an extra option which (sigh) explicitly requests them. And then make the plugin launcher read that from an env var.
Deal?

dougmassay:
1) I simply don't want to have to configure wrapper.py before hand--for the reason I mentioned: it needs to be able to be yanked out with the rest of the plugin framework--lock, stock, and barrel--and be able to be plugged into an existing Sigil install (or a plugin debug environment). I don't want to have to worry about needing to manually expand those build-time variables whenever I decide to do just that. And keep in mind that many people don't need to patch or delete anything in order to build Sigil. ;) 

2) SIGIL_EXTRA_ROOT is for relocating the entire <SIGIL_SHARE_PREFIX>/share/sigil directory (aka SIGIL_SHARE_ROOT) after Sigil has been built/installed. It's a bit of a legacy sore spot with me and quite frankly, I don't know if it's going to be around forever. I don't want to use it for anything other than relocating (and finding) the SIGIL_SHARE_ROOT directory post-sigil-installation just now. It certainly can't be used to eliminate the SIGIL_SHARE_PREFIX logic in SpellCheck.cpp like you've done. Stick to the functionality you wish to gain and don't worry so much about "improving" things.

3) I'll compromise by using USE_SYSTEM_LIBS to NOT install the bundled dictionaries, but I still don't want any hard-coded system hunspell paths in the source tree (nor a wrapper.py that needs configuring). Just add a SYSTEM_HYPHEN DICT env var to the launch script (if you're still looking to incorporate the system hyphen dictionaries) and append it's value to apaths (in wrapper.py) at the same time the value of SIGIL_DICTIONARIES (aka system_hunspell_dicts) is appended to it and be done with it. No need to rework much of anything.

eli-schwartz:
1) Yes, I said I understand that need, so I resolved to fix that.

2) I didn't (originally) eliminate the SIGIL_SHARE_ROOT logic, I simply moved it into src/CMakeLists.txt :confused: 
(So, if you use the system hunspell dictionaries, that gets compiled into Sigil, otherwise it inserts the old value directly.)
That only got rid of the env var. (Hence the fancy. Either I configured it into the executable or into the launch script, and my first approach rendered the env var obsolete again.)

When you said the launcher needs to stay separate, **I took the easy (lazy) way out** and just rebased and re-added the code to the plugin wrapper.py


I guess I will think about this in the morning then.

PR #152, opened 2015-11-07
preserve last opened file name

varlog00:
I was elsewhere doing something different. Now I do some books again. Hence this.
This patch was rejected by user none for his design(?) reasons.
So here it is again: let Sigil remember last read file name - even though it comes from plugin - and offer it as a file name when "Save" or "Save As" is invoked.

### DIFF
--- a/src/MainUI/MainWindow.cpp
+++ b/src/MainUI/MainWindow.cpp
@@ -769,6 +769,8 @@ bool MainWindow::SaveAs()
 
     // Store the folder the user saved to
     m_LastFolderOpen = QFileInfo(filename).absolutePath();
+    // Store new filename
+    m_CurrentFileName=QFileInfo(filename).fileName();
     bool save_result = SaveFile(filename);
 
     if (!save_result) {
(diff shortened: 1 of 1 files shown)


varlog00:
Would be nice to know the reasons.

kevinhendricks:
As user_none said ... the plugin name represents a temp file and is not meaningful as designed by the plugin interface.  The change is unneeded in general as you seem to be the only person pushing for it, so it adds no real value for the general plugin user.  And finally your track record on projects with me (lack of being upfront and professional) is not something that that endears your arguments in any way.

varlog00:
Thanks for answer. 
The usage case is straightforward enough: when you have to convert more then three books in the row (amazon freebies for instance) some reasonable automatic in naming files would be welcomed. Temp file could have meaningful name if preserved from original file. My slightly modified  KindleImport plugin does precisely that. I would have proposed respective changes to DiapDealer  if it made sense - which at the moment of course does not. 

How can I be professional when I'm not? It is hobby and I don't have much time for it. I learn C++/python as I work on Sigil and I'm far far away from being proficient.
I try to be upfront though - sorry if I somehow have made other impression.
Anyway, I'm glad you work on Sigil and not complaining.
Have you looked at the diff? it is such a noninvasive change?:)
OK, OK no more trolling.
Regards

PR #153, opened 2015-11-11
Compress whitespace in header text in TOC generation.

dougmassay:
See what you think of this Kevin.

I can't see any reason for line-feeds inside of header-text (generated by pretty-print) to be displayed/included in:

    1) The table of contents widget
    2) the header selection dialog when generating a ToC
    3) the navLabel text entry in the NCX file

So I've compressed whitespace in the header text in three locations using Qt's QString::simplified().

I was looking for one single place where I could compress the whitespace when the header texts are first rounded up, but I couldn't seem to find where to do it.

Anyway ... see if this even makes sense to you (and make sure it doesn't break anything else) and get back to me on it.


kevinhendricks:
Hi Doug,
Yes, that will definitely help prevent issues in the future, toc, ncx, and even the nav when we get one.  Please go ahead and merge it.

Nice job!

KevinH

dougmassay:
Thanks. Will do.

PR #160, opened 2015-11-19
make sigil build reproducibly

mapreri:
In Debian we're working on [Reproducible Builds](https://reproducible-builds.org/), and we noticed sigil couldn't reliable build reproducibly, due to a cmake globbing, which doesn't assure a stable order.
This made the linking phase unreproducible.

See also the [debian bug](https://bugs.debian.org/805500) about this.

FYI there is also another problem wrt reproducibility in sigil, due to `__DATE__` and `__TIME__`, you can see the patch in the debian sources, but it's probably too much invasive/destructive for you, that's why I'm holding myself on proposing it; can you confirm my fears?

Thanks for considering.


dougmassay:
I can't speak to the DATE/TIME issue, as I've not had time to investigate the proposed patch, but this pull request seems reasonable. The building of globbed sources is used in quite a few other places in Sigil's build process, though. Is the problem with globbed sources limited to libsigilgumbo because it's the only external library built/linked by Sigil?

mapreri:
No, this is the only place where it's needed because it's the only glob that has something to do with linking:

```
mattia@chase ~/devel/sigil/sigil (git)-[master] % find . -name CMakeLists.txt|xargs grep -i 'GLOB'
./src/CMakeLists.txt:file( GLOB TS_FILES Resource_Files/ts/sigil_*.ts )
./src/CMakeLists.txt:file( GLOB DIC_FILES Resource_Files/dictionaries/* )
./src/CMakeLists.txt:file( GLOB PLUGIN_FILES_PYTHON Resource_Files/plugin_launchers/python/* )
./src/CMakeLists.txt:file( GLOB EXAMPLE_FILES Resource_Files/examples/* )
./internal/gumbo/CMakeLists.txt:file( GLOB SOURCES *.c )
```

what are you referring with "The building of globbed sources is used in quite a few other places in Sigil's build process" ? Did I miss something?  It can be, since this particular issue is not even deterministic on its own...

dougmassay:
My mistake. I was either thinking of something in the old xerces/boost/tidy days, or confusing the Sigil source grouping with globbing.

dougmassay:
Kevin, where are we at with the translation files and freezing the code for 0.9.1? Will merging this pull request affect any of that? I wouldn't think so, but wanted to run it by you first.

kevinhendricks:
Go ahead and merge that one liner.  For the record, I think the entire idea is for the birds.  I read their justification website and even that makes no sense.  As for more invasive patches to remove date and time everywhere, where exactly are they located.  If this is just being done to make binary compares, the answer would be no.

mapreri:
On Thu, Nov 19, 2015 at 02:31:51PM -0800, Kevin Hendricks wrote:
> As for more invasive patches to remove date and time everywhere,
> where exactly are they located.

https://sources.debian.net/src/sigil/unstable/debian/patches/reproducible/

but I'm really fine if you don't accept it, I figured it.

> If this is just being done to make binary compares, the answer would be no.

This is really what the whole thing is about...

-- 
regards,
                        Mattia Rizzolo

GPG Key: 66AE 2B4A FCCF 3F52 DA18  4D18 4B04 3FCD B944 4540      .''`.
more about me:  http://mapreri.org                              : :'  :
Launchpad user: https://launchpad.net/~mapreri                  `. `'`
Debian QA page: https://qa.debian.org/developer.php?login=mattia  `-

kevinhendricks:
That patch would be removing the exact string that would tell us (and the general user) if they are using an official build of ours (matching version, date, and time) as opposed to someone else's build.  And all to let some binary comparison work, that would fail anyway if different compilers are used (clang vs gcc), different optimization levels, different debug symbol types, different compiler releases, different assemblers, different linkers, etc.  In other words this would *only* be useful for a single specific version of all build tools, switches, and distribution.

Sorry, that truly makes no sense.

jpmorgan58:
It generally makes more sense to use a non-binary comparison tool that can ignore all the unimportant metadata rather than eliminating it.

 - John

kevinhendricks:
Just a head’s up — we are about 3 or 4 days away from tagging for Sigil-0.9.1.  We are just waiting to give the transifex translation teams time to update any of the translations they desire.

So you might want to hold off for a short bit and package Sigil-0.9.1 instead  (hopefully current master - unless a serious bug report comes in) as it has a number of bug fixes important to users.  If this release proves very stable, we will branch it use that branch for all 0.9.x series of stable releases and continue active development in master.

At least that is the plan now.

mapreri:
On Fri, Nov 20, 2015 at 07:55:00AM -0800, Kevin Hendricks wrote:
> we are about 3 or 4 days away from tagging for Sigil-0.9.1.

cool!

> So you might want to hold off for a short bit and package Sigil-0.9.1
> instead

if this is for me, well, "instead" ? 0.9.0 is already in the archive :)

PR #161, opened 2015-11-21
Switch TempFolder to QTemporaryDir

pinotree:
Refactor `TempFolder` to use `QTemporaryDir` internally, instead of manually creating directories ourselves (which is not race-free). `QTemporaryDir` gives us a race-free temporary directory, which is also cleaned automatically (albeit synchronously) on `~QTemporaryDir`.

The extra hierarchy (`Sigil/scratchpad`) in the system temporary directory is no more used, as the result now is unique enough, and not shared by all the users on the same machine. As a consequence, there is no more need to fix the permissions of the scratchpad.

### DIFF
--- a/src/Misc/TempFolder.cpp
+++ b/src/Misc/TempFolder.cpp
@@ -19,71 +19,46 @@
 **
 *************************************************************************/
 
-#include <QtCore/QtCore>
 #include <QtCore/QDir>
-#include <QtCore/QFileInfo>
 #include <QtConcurrent/QtConcurrent>
 
 #include "Misc/TempFolder.h"
-#include "Misc/Utility.h"
-
-const QString PATH_SUFFIX = "/Sigil";
 
 TempFolder::TempFolder()
-    : m_PathToFolder(GetNewTempFolderPath())
+    : m_tempDir(GetNewTempFolderTemplate())
 {
-    QDir folder(m_PathToFolder);
-    folder.mkpath(folder.absolutePath());
+    // will be cleaned manually in the destructor
+    m_tempDir.setAutoRemove(false);
 }
 
 TempFolder::~TempFolder()
 {
-    QtConcurrent::run(DeleteFolderAndFiles, m_PathToFolder);
+    QtConcurrent::run(DeleteFolderAndFiles, m_tempDir.path());
 }
 
 
 QString TempFolder::GetPath()
 {
-    return m_PathToFolder;
+    return m_tempDir.path();
 }
 
 
 QString TempFolder::GetPathToSigilScratchpad()
 {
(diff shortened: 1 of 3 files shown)


kevinhendricks:
Our master is locked now in anticipation of a Sigil-0.9.1 release in the next week or so.  Once that has been accomplished, we will consider your pull request.

Thanks,
Kevin

kevinhendricks:
I am looking at merging this now but had a few questions:

- Does QTemporaryDir clean itself up in the destructor using threads?  If not, we should probably still do the cleaning up ourselves.  Unzipped epubs can get quite large and closing one window/ebook while opening the next can be a performance issue if we have to sit and wait for the clean up before leaving the destructor.

- In the past when debugging Sigil issues, it was always easy to find the "scratchpad" directory in the /tmp folder (or /private/tmp) and actually examine the internal state of the ebook files as everything related to that ebook was all in that one scratchpad directory.  Given that main.cpp no longer creates the scratchpad directory upon first boot, will we end up having to search all over the place just to find the files when debugging issues? 

- What is the primary benefit of all of these changes?  Is it just to prevent the potential for an attack via a "race" to guess the intended directory name? Doesn't generating a random uuid and appending it to the path pretty much prevent the potential for any race in guessing the intended target of temporary directory being created or am I missing something?

Thanks

pinotree:
> Does QTemporaryDir clean itself up in the destructor using threads?

No, but that can be readded back then, in a much simplier way than current way.

> In the past when debugging Sigil issues, it was always easy to find the "scratchpad" directory in the /tmp folder (or /private/tmp)

The "scratchpad" directory is a subdirectory of the system temporary directory, and now the temporary directories are just subdirectories of the system temp. With this change you find e.g. `/tmp/Sigil-XXXXXX` directories.

> Is it just to prevent the potential for an attack via a "race" to guess the intended directory name?

Yes, exactly; see http://www.tldp.org/HOWTO/Secure-Programs-HOWTO/avoid-race.html for example. Using a generic API for creating directories is susceptible to TOCTOU issue; also, the resulting directories have 0644 as permissions, so any user can see the ebooks opened by other users (and I could argue this is a security issue).

As benefit, using a standard (and well tested) Qt class makes the code easier, and safe.

kevinhendricks:
So please revise your pull request to add back in a separate thread doing the cleanup via the destructor and I'll merge it.  I like the QtConcurrent approach used now but if know of something similar that works crossplatform, I'm okay with that as well.

Thanks!

pinotree:
Pull request amended.

kevinhendricks:
Thank you for helping to make Sigil better.  Your pull request has now been merged.

PR #186, opened 2016-02-13
Added some missing MARC relators

### LINKED ISSUE #185: MARC relator list update/refresh
Now that the MARC relator list has been factored out into a separate class, it might be a good time to update it so that it conforms to the current authoritative list maintained by the LOC.

The canonical URL for the complete list is: http://www.loc.gov/marc/relators/relaterm.html

This refers to both the codes and terms as such (for example, the current MARC list no longer has a "Recipient" term and mentions that one should use the "Addressee" term instead but using the same code, `rcp`) as well as the descriptions. There are also missing terms, for example "Appellant", "Appellee", etc.

`MarcRelators::SetRelatorsMap()` mentions `    // These descriptions fixed, standard EPUB descriptions and should not be changed` , but that's misleading. The descriptions seem to come from the OPF 2 spec at http://www.idpf.org/epub/20/spec/OPF_2.0.1_draft.htm#Section2.2.6 , but that section explicitly states (emphasis mine):

> For convenience, **some** relator code values are listed here **as examples**. Consult the MARC code list cited above for the complete list.

The spec-listed descriptions themselves are in no way normative; instead, the MARC relator list definitions are. The examples aren't even being listed any more in the 3.0.1 spec: http://www.idpf.org/epub/301/spec/epub-publications.html#sec-meta-elem

The existing descriptions (i.e., term definitions) also all start with "Use for ", which is redundant in this context. The definitions on the MARC site have evolved as well to cover more use cases--for example, some list "family" in addition to "person or organization".

I am aware that changing the descriptions has implications for the translations, though. If that's undesirable for the moment, we could leave the existing ones be and just add missing relator terms using the definition from the MARC site, and see how it goes.

I volunteer to write, test and submit a patch for this (it's fairly trivial), but wanted to first sound people out on whether it would be welcome :)

mihailim:
Partial implementation of #185 (contains only *additions* to the map.) Changes left undone from the #185 proposals:

- The following terms have been renamed on the MARC relator list, keeping the relator code unchanged. The map has *not* yet been changed to reflect these changes.
  - "Recipient" -> "Addressee"
  - "Author of screenplay, etc." -> "Screenwriter"
  - "Curator of an exhibition" -> "Curator"
  - "Degree grantor" -> "Degree granting institution"
  - "Organizer of meeting" -> "Organizer"
  - "Redactor" -> "Redaktor" (although this could be a typo on LOC's list)
  - "Wood-engraver" -> "Wood engraver" (this could be a typo in Sigil's list)
- The following terms have been retired from the MARC relator list in favor of another existing term. The corresponding relator code no longer exists in the MARC relator list, so perhaps codes in edited publications should be autoreplaced? The map has *not* yet been changed to reflect these changes.
  - "Collaborator" clb (use "Contributor" ctb)
  - "Vocalist" voc (use "Singer" sng)
- The existing descriptions in the map have not been replaced with the definitions from the MARC relator list, pending advice on how to handle existing translations.


kevinhendricks:
Hi,
Would you please modify your pull to remove any obsolete relator codes as well? And I will merge your adjusted pull.  As you noted,  Redaktor is not a word so no change there.  And I am happy to *not* touch existing translations unless absolutely needed.  Most people will be key off the word then only look at definition at the bottom for confirmation so repetitive phrases should not be a huge issue making keeping the translation more important.

This list is not used to validate anything and the user can of course use anything they wish, so I am not worried about older unused codes in current metadata.  I would rather keep the list more manageable in size.

Thanks

mihailim:
Thanks for looking at this! Done, dropped the obsolete codes.

kevinhendricks:
thank you

PR #187, opened 2016-02-13
Updated Transifex project URL

mihailim:
The existing Transifex URL in Translating.md is outdated and leads to hxxps://www.transifex.net which uses a certificate expired in 2013 and then redirects to the current URL. Updated to directly point to the correct Transifex project URL.


kevinhendricks:
Thank you.

PR #194, opened 2016-03-15
Modify W3C CSS validation tool

dougmassay:
EPUB3 always validates using CSS 3.0. EPUB2 validates using user choice (CSS 2.1 or CSS 3.0) in General preferences (defaults to CSS 2.1). SettingsStore entries are in place to make EPUB3 validation a user preference as well, but is unimplemented currently.


dougmassay:
Do you think it's overkill to provide different preferences for EPUB2 VS EPUB3? perhaps a single CSS validation preference affecting both EPUB3 and EPUB2 will suffice here? Pretty easy to go either way from here.

PR #212, opened 2016-04-25
Typo in href_to_basename

wrCisco:
This is simply a typo, I think. With 'href' the function works properly.


kevinhendricks:
yes, nice catch.
Thank you.

PR #222, opened 2016-05-28
The format is spelled "EPUB" according to http://idpf.org/epub

probonopd:


kevinhendricks:
Google epub.
You will find all versions of the acronym all using different case.  Including official publishing related documents and websites. It was even that way on the original idpf website for epub2.  Our own Mobileread forum is called "ePub".  Just because the idpf site changed to standardize on EPUB does not mean the world has or ever will as an Electronic Publishing  or Electronic Publication can be abbreviated in many ways

So no thank you.

PR #234, opened 2016-08-02
Instruct sigil to look up a couple of path by default on linux for dictionaries

mapreri:
add /usr/share/hunspell/ and /usr/share/myspell/ to the used paths.

Those are the paths where spelling dictionaries are installed on most linux
distributions.

Used only when neither SIGIL_DICTIONARIES nor SIGIL_EXTRA_ROOT are set.

Would you be open to this kind of change?
The aim is to drop the patch in the loader that is just setting SIGIL_DICTIONARIES, when sigil could look up that path by itself.  FYI, libreoffice defaults on /usr/share/hunspell (and falls back to /usr/share/myspell if the former doesn't exists) and it's used on debian and derivatives (where we are deprecating /usr/share/myspell); whereas /usr/share/myspell is used on redhat and derivatives, and apparently /u/s/hunspell doesn't exist anywhere.  Also scribus started to look up these paths by itself in the recent 1.5.2 development release.

### DIFF
--- a/src/Misc/SpellCheck.cpp
+++ b/src/Misc/SpellCheck.cpp
@@ -342,11 +342,16 @@ void SpellCheck::loadDictionaryNames()
         }
     }
     // else use the env var runtime overridden 'share/sigil/hunspell_dictionaries/' location.
-    else if (!sigil_extra_root.isEmpty()) {
-        paths.append(sigil_extra_root + "/hunspell_dictionaries/");
-    } else {
-        // else use the standard build time 'share/sigil/hunspell_dictionaries/'location.
-        paths.append(sigil_share_root + "/hunspell_dictionaries/");
+    else {
+        if (!sigil_extra_root.isEmpty()) {
+            paths.append(sigil_extra_root + "/hunspell_dictionaries/");
+        } else {
+            // prepend the usual places where hunspell dictionaries are located in unix systems
+            paths.append("/usr/share/myspell/");
+            paths.append("/usr/share/hunspell/");
+            // else use the standard build time 'share/sigil/hunspell_dictionaries/'location.
+            paths.append(sigil_share_root + "/hunspell_dictionaries/");
+        }
     }
 #endif
     // Add the user dictionary directory last because anything in here
--- a/src/Resource_Files/plugin_launchers/python/wrapper.py
+++ b/src/Resource_Files/plugin_launchers/python/wrapper.py
@@ -828,11 +828,14 @@ class Wrapper(object):
                 if len(share_override):
                     apaths.append(unipath.abspath(os.path.join(share_override, "hunspell_dictionaries")))
                 else:
+                    # prepend usual places where hunspell dictionaries are located in unix systems
+                    apaths.append(os.path.join(os.sep, 'usr', 'share', 'myspell')
+                    apaths.append(os.path.join(os.sep, 'usr', 'share', 'hunspell')
                     # Otherwise, use Sigil's bundled hunspell dictionary location.
                     apaths.append(unipath.abspath(os.path.join(share_prefix, "share", 'sigil', "hunspell_dictionaries")))
             apaths.append(unipath.abspath(os.path.join(self.usrsupdir,"hunspell_dictionaries")))
         return apaths
-                        
+
     def get_gumbo_path(self):
         if sys.platform.startswith('darwin'):
             lib_dir = unipath.abspath(os.path.join(self.appdir,"..","lib"))
@@ -844,7 +847,7 @@ class Wrapper(object):
             lib_dir = unipath.abspath(self.appdir)
             lib_name = 'libsigilgumbo.so'
         return os.path.join(lib_dir, lib_name)
-            
+
     def get_hunspell_path(self):
         if sys.platform.startswith('darwin'):
             lib_dir = unipath.abspath(os.path.join(self.appdir,"..","lib"))


mapreri:
while doing this I changed the indentation and added a couple of brackets.  If you want I can split that in 2 commits.

dougmassay:
I'll check it out when I get home later. Suffice to say that I'm not "opposed." But any such change also has to work with Python plugin framework--which also has its own logic to find the appropriate hunspell dictionaries. Your merge request doesn't account for that aspect.

I would also prefer that the standard system dictionary locations were used if--and only if--the bundled dictionaries were not present in Sigil's share directory (wherever it may be). In other words... I wouldn't want any installed system hunspell dictionaries being prefered (just because they were first in the search path) when a user has built Sigil with the default option of installing the bundled dictionaries.

In short ... if the bundled dictionaries are present in Sigil's support (share) directory (wherever it may be, or be relocated to at runtime), then they need to be the ones being used. Both by Sigil and by Sigil's Python plugin framework. Make that happen and I'll have no philosophical objections. ;)

mapreri:
Right.. I also noticed the python plugins while grepping for
SIGIL_DICTIONARIES, but then forgot about it :/
Sure I can work on making the sigil-installed the defaults, though it's
gonna take a little while (=> not in the next few hours) :)

Thanks for being "not opposed"!!

dougmassay:
No problem. Even if you simply make the sysyem directories the last paths appended, it should be fine. If Sigil is built with USE_BUNDLED_DICTS=0 (as all repo builds will be), then there won't be any hunspell dictionaries in those first locations anyway.

mapreri:
> I would also prefer that the standard system dictionary locations were used if--and only if--the bundled dictionaries were not present in Sigil's share directory (wherever it may be). In other words... I wouldn't want any installed system hunspell dictionaries being prefered (just because they were first in the search path) when a user has built Sigil with the default option of installing the bundled dictionaries.

besides, are you sure of this?  in src/Misc/SpellCheck.cpp there is this comment:
```
    // Add the user dictionary directory last because anything in here
    // will override installation supplied dictionaries.
```

dougmassay:
Sorry. I wasn't including any manually-added user-supplied dictionaries (which are located in the user's ./local preference directories). I just meant that I don't want any repo installed dictionaries trumping the bundled dictionaries, is all. If a user custom compiled Sigil with the default options, and then later installs some repository dictionaries, Sigil should still continue to use the bundled dictionaries. Unless they choose to override withe env var.

Just append your /usr/share/hunspell and /usr/share/myspell directories AFTER the sigil_share_root + "hunspell_dictionaries" entry (instead of before) and all should be well. The latter won't have any dictionaries in it on a repo-built Sigil anyway. Because it's built with the INSTALL_BUNDLED_DICTS=0 option. Therefore, the repo-built Sigil will use the dictionaries in the last-chance /usr/share/hunspell location (again, unless the user overrides with the env var).

mapreri:
What I am saying is that if that comment is true, then appending my entries
(as opposed to prepending as I did) would do exactly what you are opposed
to.

dougmassay:
Ah... I see what you're saying now. I didn't have the whole file in front of me until now.

I don't like the idea of lumping any potentially installed system dictionaries in with any potentially installed bundled dictionaries. It's certainly fine as is when looking at it strictly from a repo-installed Sigil angle (since there'll be no bundled dictionaries to worry about anyway). 

But it complicates things when there's system dictionaries present and a user manually compiles Sigil with the bundled dictionary option. 

There is, right now, a clear "Location A or Location B or Location C (plus the user_dictionary directory)" scenario. What you're proposing would essentially be a: "Location A or Location B or Locations C & D (plus the user_dictionary directory)" scenario. That troubles me a bit, I would rather there be a A or B or C or D (plus the user_dictionary directory) solution if at all possible.

eli-schwartz:
In #139 I made it a compile-time choice, and added `SIGIL_DICTIONARIES` to the variables configured into the launcher script in order to maintain compatibility with the plugin framework.

I am not entirely sure why that PR was rejected without comment, *after* I fixed the issues @dougmassay raised.
It was doing the right thing in all circumstances...

dougmassay:
> I am not entirely sure why that PR was rejected without comment, after I fixed the issues @dougmassay raised.
It was doing the right thing in all circumstances...

I was pretty clear why I rejected your proposal at the time, I think. And while you did eventually "fix" all the issues I raised, it was unfortunately only after I was already working on a compromise I liked better. Apologies for not commenting on the closure.

I won't make the same mistake here. ;)

I'm working on a plan that will hopefully meet with everyone's approval. No more patching. The system dict directories will be able to specified at compile time (or at run time), and I'll be able to eliminate the ugly duplicated logic of selecting the correct hunspell dictionary directories all over again in the Python plugin framework.

dougmassay:
@mapreri:

I'm going to accept this merge request to get the ball rolling, but I wanted to warn you that some of it is going to be refactored a bit (something I've been meaning to do regardless of this PR).

The main functional difference will be that only the /usr/share/hunspell directory is going to be hardcoded/added automatically (whenever -DINSTALL_BUNDLED_DICTS=0 is used at compile time). If you want additional paths to be included, you'll be able to add them using the new EXTRA_DICT_DIRS compile time option (a colon-separated list of paths). A la -DEXTRA_DICT_DIRS=/usr/share/myspell:/usr/share/yourspell.

I hope this is acceptable.

eli-schwartz:
> I was pretty clear why I rejected your proposal at the time, I think. And while you did eventually "fix" all the issues I raised, it was unfortunately only after I was already working on a compromise I liked better. Apologies for not commenting on the closure.

Thanks for clarifying. I suspected, but was not sure, that that might be the case.

> I'm working on a plan that will hopefully meet with everyone's approval. No more patching. The system dict directories will be able to specified at compile time (or overridden at run time), and I'll be able to eliminate the ugly duplicated logic of selecting the correct hunspell dictionary directories all over again in the Python plugin framework.

That sounds good. Although I imagine the only need to specify the system dictionaries is at compile time, to conform to the system layout. Any user-installed extra dictionaries in non-distro-defined locations should go in the Sigil prefs dir.

dougmassay:
> Although I imagine the only need to specify the system dictionaries is at compile time, to conform to the system layout. Any user-installed extra dictionaries in non-distro-defined locations should go in the Sigil prefs dir.

Absolutely. Any additional system dictionary paths provided via the  -DEXTRA_DICT_DIRS option will only be relevant when -DINSTALL_BUNDLED_DICTS=0 is in play.

mapreri:
> The main functional difference will be that only the /usr/share/hunspell directory is going to be hardcoded/added automatically (whenever -DINSTALL_BUNDLED_DICTS=0 is used at compile time). If you want additional paths to be included, you'll be able to add them using the new EXTRA_DICT_DIRS compile time option (a colon-separated list of paths). A la -DEXTRA_DICT_DIRS=/usr/share/myspell:/usr/share/yourspell.

this sounds just good for my use case, thanks! ♥

> > Although I imagine the only need to specify the system dictionaries is at compile time, to conform to the system layout. Any user-installed extra dictionaries in non-distro-defined locations should go in the Sigil prefs dir.

> Absolutely. Any additional system dictionary paths provided via the -DEXTRA_DICT_DIRS option will only be relevant when -DINSTALL_BUNDLED_DICTS=0 is in play.

note that there @eli-schwartz was talking about *user* dictionaries, not system's, that are currently expected to be in ~/.local/share/sigil-ebook/sigil/hunspell_dictionaries and always added to the search path, and I assume this won't change.

dougmassay:
> note that there @eli-schwartz was talking about user dictionaries, not system's, that are currently expected to be in ~/.local/share/sigil-ebook/sigil/hunspell_dictionaries and always added to the search path, and I assume this won't change.

That remains the same, yes.

dougmassay:
I refactored the solution after merging as discussed. Please ensure that this is still working as planned. Thanks for the contribution.

eli-schwartz:
It works as intended, so I just pushed a new sigil-git to the [AUR](https://aur.archlinux.org/cgit/aur.git/coAURmmit/?h=sigil-git&id=20f4057c40b90e59096d19e6c40c0ae5397222ed) that no longer needs to patch the launch script.

dougmassay:
Good to know. Thanks for the followup. Just note that there's a few new things in HEAD right now that may not be completely "ready for primetime" yet. Everything should still certainly _work_ but there are some things that are in flux right now. Don't you typically pull a Release tag from git for your AUR package?

mapreri:
Yeah!  I uploaded to Debian too :smile: (but instead I just cherry-picked the 2 last commits fb05355ef0441a75234f6f836159de68f461a6a7 and 8e6b1538b6de391bbdf2197ae81f6894eb051244)

Thanks a lot @dougmassay !

kevinhendricks:
@eli-schwartz @mapreri 
While we are at it, are their any other outstanding linux distribution specific patches that are needed for building distribution versions of Sigil that we should be incorporating?

eli-schwartz:
> Good to know. Thanks for the followup. Just note that there's a few new things in HEAD right now that may not be completely "ready for primetime" yet. Everything should still certainly work but there are some things that are in flux right now. Don't you typically pull a Release tag from git for your AUR package?

The release tag is used by the stable release in the community repositories. The AUR package "sigil-git" is, appropriately, the development preview. :wink: 

> While we are at it, are their any other outstanding linux distribution specific patches that are needed for building distribution versions of Sigil that we should be incorporating?

I can't speak for other distributions, but the system dictionaries was the only patch used anywhere in Arch Linux (official repos or AUR).

mapreri:
@kevinhendricks 
> While we are at it, are their any other outstanding linux distribution specific patches that are needed for building distribution versions of Sigil that we should be incorporating?

As you know, in debian we are pretty strict on not including built stuff (minified js counts as built stuff), and also on embedding 3rd party libs.
As such I have [this patch](https://anonscm.debian.org/git/collab-maint/sigil.git/tree/debian/patches/mathajx?h=debian/0.9.6%2bdfsg-2) to be able to use regular mathjax (which is not single-bundled but needs to be configured at loading time) + I remove all files which are listed in the Files-Excluded field of [debian/copyright](https://anonscm.debian.org/git/collab-maint/sigil.git/tree/debian/copyright?h=debian/0.9.6%2bdfsg-2).

I think removing `3rdparty/` stuff is ok from my part (and also kind of supported by you).  Possibly replacing jquery and jquery.scrollto by system-wide ones might not be (but it's working great, even if your embedded versions are quite old even compared to what we have in the debian archive).

I definitely wanted to think of a way of "providing" you some kind of upstreamable solution to avoid having to carry the mathjax patch, but at least as of now I can't think of anything sane for that.

kevinhendricks:
Having an external normal (non-single)  mathjax with working configuration is not a bad thing.  I personally like the single version as it loads faster.  That said, I would be happy to accept a pull request that used a CMake define to set the path to an external mathjax and also added an ifdef to control the extra code you need as long as it doesn't break our Linux, Mac and Windows use.

Also, what versions of jquery and jquery-scroll-to are you using?   I have been meaning to update these to something much more recent but did not want to risk breaking the mess of js that is used in Codeview, Bookview, and Preview. Given a more recent version actually works for you, I should be able to update the earlier versions to whatever you are using safely.

Pull requests that move these minified js to external and that do NOT break the current Linux, Mac, and Windows build and operation also happily accepted if you do not like the symlink approach you are using.

FYI, compiling with XCode 7.2 clang 64 bits yesterday, reveals undefined behaviour shifting negative values in a few of the 3rd party (external) libs.  We are going to patch and update our current versions to prevent that compiler issue.  Not sure if they still exist in your versions, but thought someone might like to know.

Kevin

mapreri:
> Also, what versions of jquery and jquery-scroll-to are you using?

jquery 1.12
jquery-scroll-to 2.1.2

> I have been meaning to update these to something much more recent but did not want to risk breaking the mess of js that is used in Codeview, Bookview, and Preview. Given a more recent version actually works for you, I should be able to update the earlier versions to whatever you are using safely.

well, I don't use every single feature of sigil myself...  e.g.  I never use preview myself.
though, this set of jquery/jqyery-scroll-to is there since quite some time, and nobody yet complained of them.

> Pull requests that move these minified js to external and that do NOT break the current Linux, Mac, and Windows build and operation also happily accepted if you do not like the symlink approach you are using.

oh, I can live with the symlinks pretty well.  Just figured I'd mention it given that you asked.
though if I'm always happy to remove local stuff and plain the differences with upstreams' defaults.

> FYI, compiling with XCode 7.2 clang 64 bits yesterday, reveals undefined behaviour shifting negative values in a few of the 3rd party (external) libs. We are going to patch and update our current versions to prevent that compiler issue. Not sure if they still exist in your versions, but thought someone might like to know.

possibly.  doing dynamic linking also means I don't have to care about libaries building :)
you can easily see the version of the libs debian builds against in https://packages.debian.org/sid/sigil

PR #237, opened 2016-08-26
Fix UI language setting resetting back to English

bbacskay:
When the language code contains "-". The setting is saved with "-" replaced to "_", but this was not changed back when dialog was loaded.


kevinhendricks:
Yes, with the underscore it won't be found in the language list which uses dadhes.

Nice Catch!

Thank you!

PR #244, opened 2016-10-09
first draft of Settings and Preferances, numerous tweaks

varlog00:
The main functionality is there. I'll be cleaning the source and looking for missed items next, but I would like to have some input.
The Nice:
it works on my system
The Ugly:
the way SpellCheck remembers its last session
double parsing by CodeEditor and Highlighter
and others

### DIFF
--- a/src/Dialogs/PreferenceWidgets/SpellCheckWidget.cpp
+++ b/src/Dialogs/PreferenceWidgets/SpellCheckWidget.cpp
@@ -55,7 +55,9 @@ void SpellCheckWidget::setUpTable()
     ui.userDictList->setModel(&m_Model);
     // Make the header fill all the available space
     ui.userDictList->horizontalHeader()->setStretchLastSection(true);
+    ui.userDictList->horizontalHeader()->setDefaultAlignment(Qt::AlignLeft);
     ui.userDictList->resizeColumnToContents(0);
+    ui.userDictList->resizeColumnToContents(1);
     ui.userDictList->verticalHeader()->setVisible(false);
     ui.userDictList->setSortingEnabled(false);
     ui.userDictList->setSelectionBehavior(QAbstractItemView::SelectRows);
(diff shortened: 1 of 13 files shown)


kevinhendricks:
All merged in.  Thanks for working on this!

varlog00:
Thanks.

As I've said some input would be nice. Does it compile? Is UI agreeable?

I go through the source from the beginning now, cleaning and 
implementing missed items - which are many.

I've removed from SearchOperations.cpp function:

std::tuple<QString, int> 
SearchOperations::PerformHTMLSpellCheckReplace(const QString &text,
         const QString &search_regex,
        const QString &replacement)

seems not to be used by anything - do you know some reason to keep it?

dougmassay:
It compiles OK for me on Linux and Windows, but that's all I've had time to check.

Please don't remove any functions unless you added them yourself and no longer need them.

varlog00:
Thanks.

Just for clarification: when I write (for an instance) a function 
SpellCheck::spellML(...), which is supposed to replace original 
SpellCheck::spell(...) function, I have to remove the original 
eventually. So I will be removing functions which conform to your objection.

There is a Sigil single dictionary mode, which I'm thinking of but 
haven't started yet, where the said functions would be handy, of course 
- but I would like to get ML going first, with all Sigil's previous 
functionality at least. Original functions are still there, in git, forever.

The function in question is something so bizarre that I don't want to 
get to know the hows and whys of it - at the moment - because I'm doing 
something else (ignored words, user dictionaries - not quite trivial). 
And it seems to replicate already existing functionality. As it has no 
client (there is only definition of it there in .h and .cpp) and it 
prevents compiling, I have just commented it out - which is my general 
policy in such cases. I hope you are fine with it, considering we are 
speaking of temporary branch here.

dougmassay:
That sounds reasonable, yes. I'm just thinking of when it comes time to merge the changes from this branch into Master. A lot of extra "clean up" can sometimes actually hinder that process by creating more conflicts at merge time. It's sometimes better to leave any "tidying up" that needs done until after the branch is merged back into master.

PR #255, opened 2016-11-20
Implemented user dictionaries / ignored words, numerous tweaks 

varlog00:
Implementing user dictionaries and ignored words turned out to be quite complicated. Obviously both functionalities should be now language aware in a way they weren't before. But what happens to old dictionaries, which the users of Sigil have tended to all those past years?
In order not to break  current usage I decided on following architecture: user dictionary consists of main dictionary "name" and sub-dictionaries "name.dictionary_name" (for instance "default.en_GB" or "default.en"). The main dictionary will be applied to all used dictionaries, sub-dictionary only to appropriate language dictionary. The sub-dictionaries will be automatically created as needed. One of bad consequences of this is that main dictionary name cannot have suffix.
This is probably overkill.
Anyway, I consider the functionality of multilanguage spell check to be on the same level as SIgil's original. I will be still working on it - but mostly I will be waiting for some input. Some independent testing is long overdue. 
  
regards

### DIFF
--- a/src/Dialogs/PreferenceWidgets/SpellCheckWidget.cpp
+++ b/src/Dialogs/PreferenceWidgets/SpellCheckWidget.cpp
@@ -27,6 +27,9 @@
 #include <QtGui/QDesktopServices>
 #include <QtWidgets/QInputDialog>
 #include <QtWidgets/QMessageBox>
+#include <QtWidgets/QTreeView>
+#include <QCryptographicHash>
+#include <QDebug>
 
 #include "SpellCheckWidget.h"
 #include "Misc/Language.h"
@@ -38,7 +41,10 @@ const QString DEFAULT_DICTIONARY_NAME = "default";
 
 SpellCheckWidget::SpellCheckWidget()
     :
-    m_isDirty(false)
+    m_isDirty {false},
+    m_hash {},
+    m_dirtyDicts {},
+    m_Model {new QStandardItemModel(this)}
 {
     ui.setupUi(this);
     setUpTable();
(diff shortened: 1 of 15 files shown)


varlog00:
Sorry, was plying with github and temporary deleted the branch :(.  Pull request still stands.

kevinhendricks:
FWIW, I would not make user dictionaries language specific.  Let the user control that list.  I would verify/ignore words against user specified word not considering language.  In other words treat user word lists as if they match all languages.

Kevin

varlog00:
hmm... I'm using Multilanguage01.epub as my test case of course. It is some over 10 dictionaries and the time print is noticeable:  I've tried to account for it. It is not usual usage case, I agree. Will think about it and 
will throw my hard(!) work with QStandardItemModel away for something more practicable... business as usual... if the time and circumstances allow :).

varlog00:
but look at it as it is: would be a pity to let it die without considering...
Just after I've committed it I I've noticed some annoying bugs... Changes pending anyway...  if the time and circumstances allow :).

varlog00:
Hi Kevin,
still considering? it is a just a brunch...
The talk about preserving file name (which you promised) is concluded and everybody is happy.
Any changes I'll do will be build on top of this commit.
I will be probably a month or so away - so don't hold your breath :).
and if you don't feel compatible enough with me to continue this venture - just say so.

regards

kevinhendricks:
I still am quite tied up with work so I haven't had a chance to look in detail at anything.
I will try to do that over the Christmas holidays and then merge what I can, piece by piece.

I will look at the file name stuff for the plugin at the same time.

No worries,

Kevin

varlog00:
> I will look at the file name stuff for the plugin at the same time.

Don't bother. Everybody is happy the way it is - so I've canceled pull request and deleted branch.
The spell check, on the other hand, I'd like to bring to some final conclusion - whatever it may be :).

PR #256, opened 2016-11-20
preserve file name

varlog00:
All programs I use, when converting from one format to another, allow to keep original file name - only changing the suffix. Not so Sigil :(.
This patch brings such an ability to Sigil when using converting plugins like KindleImport - assuming KindleImport allows it by preserving this name  (I'm willing to provide necessary changes to plugin, too) .

### DIFF
--- a/src/MainUI/MainWindow.cpp
+++ b/src/MainUI/MainWindow.cpp
@@ -797,6 +797,8 @@ bool MainWindow::SaveAs()
 
     // Store the folder the user saved to
     m_LastFolderOpen = QFileInfo(filename).absolutePath();
+    //store new file name
+    m_CurrentFileName=QFileInfo(filename).fileName();
     bool save_result = SaveFile(filename);
 
     if (!save_result) {
@@ -3659,6 +3661,7 @@ void MainWindow::CreateNewBook()
     SetNewBook(new_book);
     new_book->SetModified(false);
     m_SaveACopyFilename = "";
+    m_CurrentFileName.clear();
     UpdateUiWithCurrentFile("");
 }
 
@@ -3710,6 +3713,9 @@ bool MainWindow::LoadFile(const QString &fullfilepath, bool is_internal)
             if (!m_IsInitialLoad) {
                 ShowLastOpenFileWarnings();
             }
+
+            //preserve file name
+            m_CurrentFileName=QFileInfo(fullfilepath).fileName();
 
             if (!is_internal) {
                 // Store the folder the user opened from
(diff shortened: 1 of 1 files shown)


dougmassay:
When all is said and done, all of this still relies on Input Plugin devs naming their temporary epubs with a naming scheme that works for _you_. Namely, the file name (minus the extension) of the original document (if there was, in fact, a single document for the source of the input plugin in the first place). It just seems like a lot of effort (and some awkward logic) to get right back into nearly the very same boat:  _hoping_ the input plugin dev names their temporary epub the way you want them to.

I'm not opposed to the concept of having the Save As dialog pre-populated with a sane, relevant suggestion when first saving after and input plugin runs. Not _per se_, anyway. But I'm not really in favor of a solution that's still going to require you to "educate" input plugin devs on how you believe they should be naming the temporary epub their plugin produces. All to save a little one-off typing on the very-first save.

I'm not opposed. But I'd much rather see a solution that doesn't put the burden on the plugin devs to accomplish what is -- let's be honest -- one person's personal preference.

varlog00:
Well, well, well...
Sigh...
OK.
When Input Plugin Devs name their output  "fuck.you" all is lost anyway. Sigil, as it is designed , has no means to react to it other then "please, fuck me" or  "fuck you, I do my own thing" (as far as I can see, please enlighten me if I'm wrong: Sigil can analyze the book itself, but the source data is gone).
So I propose to teach the devs not to say those inappropriate words: let them say something meaningful. and Sigil react appropriate - in this singular case.

There is at least one other person (http://www.mobileread.com/forums/showpost.php?p=3430314&postcount=9), which seems  to be in thinking somehow as I do (other context, I know). 

`All to save a little one-time typing on the very-first save`
It is a question of user comfort. In this case I' m the user. I would like just to click OK. All known to me professional programs allow this. No, Sigil (at the moment) wants me remember WTF'ing name this amazon freebie was ASDFGGSSJJSDS.this or WJJJDJHHDHDH.that. I just want to known if I've already converted it.

`(and some awkward logic) `
please, elaborate.

dougmassay:
I tried. But it's clear you're unable to converse rationally and politely about this.

varlog00:
If word f...c offends you I humbly apologize. I will not use it anymore - my peculiar sense of humor and perhaps my incomplete grasp of appropriate idiomatic English is guilty.

varlog00:
OK, doesn't really matter. 
Was just looking for some user (me :)) comfort... and trying to correct things which I consider fundamentally  (really, that's how I see it and that's why I've tried to change it) wrong in a program I **like**.  
I'm using patched Sigil and KindleImport anyway, will just have to go through the hassle by new versions, no big deal.

thanks for all the fish ...

PR #278, opened 2017-02-04
Write EPUB everywhere in interface

bershan2:
In src/Resource_Files/ts/base.ts replace all occurrences of "Epub" in interface text with "EPUB", which is the way IDPF writes it.
Translations left unchanged: since they will be pulled from transifex.com, "EPUB" correction has to be done there.

### DIFF
--- a/src/Resource_Files/ts/base.ts
+++ b/src/Resource_Files/ts/base.ts
@@ -1970,18 +1970,18 @@ other end to continue searching.</source>
     </message>
     <message>
         <location filename="../../Form_Files/PGeneralSettingsWidget.ui" line="20"/>
-        <source>Choose which version of Epub to use
-when creating new or empty Epubs in Sigil.</source>
+        <source>Choose which version of EPUB to use
+when creating new or empty EPUBs in Sigil.</source>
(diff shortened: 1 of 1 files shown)


kevinhendricks:
Sorry. No interested.  The original IDPF symbol and name was ePub.  They have changed its case repeatedly.

bershan2:
Do you mean you'll reject any commit with ePub/epub/EPUB/etc. standardization in principle? Or do you dislike how "EPUB" looks? I just wish the UI was consistent and could change the case to the one you like best, if you'll accept such commit.

kevinhendricks:
yes, I have no interest in any pull requests to standardize on any single  version of epub.  

That said pull requests for real typos and not opinions are welcome

kevinhendricks:
Also, base.ts is a generated file from the set of acutal source code and nothing should be changed in it.  Instead the underlying source code the typo was in, has to be fixed.  I pushed a fix for src/Dialogs/MetaEditor.cpp to fix your "asequence" to "a sequence" in the right place since I was fixing a different typo reported via transifex issues.

PR #288, opened 2017-04-28
CodeViewEditor.cpp: PVS-Studio: CWE-476 NULL Pointer Dereference.

SvyatoslavRazmyslov:
We have found and fixed a weakness using PVS-Studio tool. PVS-Studio is a static code analyzer for C, C++ and C#: https://www.viva64.com/en/pvs-studio/

We suggests having a look at the emails, sent from @pvs-studio.com.

Analyzer warning: [V595](https://www.viva64.com/en/w/V595/) The 'end_node' pointer was utilized before it was verified against nullptr. Check lines: 2277, 2280. CodeViewEditor.cpp 2277


kevinhendricks:
Actually
 
   GumboNode* end_node = gi.get_node_from_qwebpath(webpath);

will always return a valid node even after parsing a completely empty document as it will add the enclosing html and body tags.  The only way this could fail would be an out of memory condition and then the whole thing is shot to hell anyway.

But your change can not actually hurt anything.

Merging, thank you.

PR #289, opened 2017-04-28
PVS-Studio: An exception should be caught by reference rather than by value.

SvyatoslavRazmyslov:
We have found and fixed a bug using PVS-Studio tool. PVS-Studio is a static code analyzer for C, C++ and C#: https://www.viva64.com/en/pvs-studio/

We suggests having a look at the emails, sent from @pvs-studio.com.

Analyzer warnings: 

- V746 Type slicing. An exception should be caught by reference rather than by value. main.cpp 334
- V746 Type slicing. An exception should be caught by reference rather than by value. MainWindow.cpp 3733
- V746 Type slicing. An exception should be caught by reference rather than by value. MainWindow.cpp 3841

### DIFF
--- a/src/MainUI/MainWindow.cpp
+++ b/src/MainUI/MainWindow.cpp
@@ -3730,7 +3730,7 @@ bool MainWindow::LoadFile(const QString &fullfilepath, bool is_internal)
         Utility::DisplayStdErrorDialog(
             tr("The creator of this file has encrypted it with DRM. "
                "Sigil cannot open such files."));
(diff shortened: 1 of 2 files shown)


kevinhendricks:
Catching by value is not a bug.  It can only be an issue if you have exception inheritance hierarchies with different exception messages or if you rethrow (it can convert the thrown class to base type).

It is convention to catch by reference so that inherited types are consistent and if exceptions are used for anything more than local error messages.

None of that is going on here as far as I can tell.

So please explain why you think these instances are actual "bugs".

PR #348, opened 2018-03-29
Not translated hints in Find & Replace window

BeckyDTP:


kevinhendricks:
Please remove any and all changes to base.ts.  That file is autogenerated during a release build from scratch before it is posted for translation just before a release.  See our docs on translation and how to rebuild your own base.ts using Qt tools.

Once you have removed that part, I would be happy to merge the remaining piece of the pull.

BeckyDTP:
I know that I can generate my own base.ts file
Now should be OK – only one files is changed (src/MainUI/FindReplace.cpp).

And what do you think [about it](https://github.com/BeckyDTP/Sigil/commit/5a31657963206e37aade5c25beb4e8b6825a1426)? 
Currently Plugins menu is without accelerator.

kevinhendricks:
Since menu items under Plugins have no keyboard accelerators (and can not as they are dynamically generated and may disappear at any time) there is no need for that change.

PR #355, opened 2018-05-21
Fix two Chinese translation errors

kevin-zqw:
1. Open Tab For Image
2. Mark Selected Text


dougmassay:
All Translations are done by Transifex volunteers. You'll have to get someone from the Sigil translation project at Transifex to make the correction there.

We don't accept pull requests for the ts files here. We just pull them in from Transifex at release time.

dougmassay:
https://www.transifex.com/zdpo/sigil/

Issue #356, opened 2018-05-21
Crash generating TOC if <h123> contains Unusual Chinese characters 


kevin-zqw:
If we use Unicode C, D, E, F region Chinese characters in h1 or h2, sigil will crash when generating the TOC.

```
TTraceback (most recent call last):
  File Sigil/python3lib/xmlprocessor.py line 185 in repairXML
    if _well_formed(newdata):

  File Sigil/python3lib/xmlprocessor.py line 70, in _well_formed
    newdata = newdata.encode('utf-8')

UnicodeEncodeError: 'utf-8' codec can't encode character '\udb80' in position 559: surrogates not allowed
```

if we add the surrogatepass option as follow:
`newdata = newdata.encode('utf-8', 'surrogatepass')`

the encode call will not crash

kevin-zqw:
#357

kevin-zqw:
will be fixed, thanks @kevinhendricks

PR #357, opened 2018-05-21
Fix utf-8 encode exception if surrogate presented

kevin-zqw:
If surrogate presented in the str, the encode('utf-8') will throw an exception and make sigil crashed. Add the surrogatepass option to allow surrogate.

### DIFF
--- a/src/Resource_Files/python3lib/xmlprocessor.py
+++ b/src/Resource_Files/python3lib/xmlprocessor.py
@@ -67,7 +67,7 @@ def _well_formed(data):
     result = True 
     newdata = data
     if isinstance(newdata, str):
-        newdata = newdata.encode('utf-8')
+        newdata = newdata.encode('utf-8', 'surrogatepass')
     try:
         parser = etree.XMLParser(encoding='utf-8', recover=False)
         tree = etree.parse(BytesIO(newdata), parser)


kevinhendricks:
Sigil requires all xhtml to uses only one encoding, utf-8.  Anything else is an encoding error.  I am confused.  Is your xhtml properly encoded at pure utf-8?  Why does Chinese need surrogate pairs (a utf-16 construct) in a utf-8 encoded xhtml file?

I would prefer that errors are ignored if need be, instead of smuggling byte codes that are not legal in utf-8 encodings.

Would you please enlighten us as to why errors should not be either replaced or ignored when processing a pure utf-8 encoded xhtml file?

Thanks,

kevin-zqw:
Hi, kevinhendricks
Thanks so much. 

There are many Chinese characters in Unicode region C, D, E, F. which we may use in h1, h2. Then when sigil try to encode the TOC file, it will crash.

I agree that we should not smuggling byte codes that are not legal in utf-8 encodings, but that's better than sigil crash.

Thanks again.

kevinhendricks:
Agreed, crashing is very bad and will be fixed in the next release but I need to understand how a properly encoded utf-8 xml file is causing the crash you are seeing, so that I can understand which error handler is appropriate (replace vs ignore vs xmlcharrefreplace vs surrogatepass) and why.

And I am still confused.  

From my understanding,  utf-8 does NOT need or use surrogates.  Using a variable number of bytes It can directly represent the entire range of characters from U+0000 to U+10FFFF.

If you have a utf-8 encoded file, there is no need or use for surrogates.   If you have a xhtml or xml file that says it is encoded in utf-8, NO surrogates can or should exist in that file, otherwise the file is improperly encoded.

If you have a utf-16 encoded file, then surrogate pairs are needed to represent any character outside the BMP.

The xmlprocessor.py code you are using is used to process xml encoded files that are stored as utf-8 encoded stream of bytes and all conversion from utf-16 is handled when the epub is first loaded into Sigil the first time.

So any time I am given a string in full unicode (in python running embedded inside Sigil) I should be able to encode it properly as utf-8 without any errors being generated and with no surrogates needed.

What am I missing?   Am I missing something or is my understanding of utf-8 and surrogates incorrect?

Please attach a zip file of the exact single xml file that causes the error so that I can see if that file is properly decoded and then encoded as utf-8 when first loaded into Sigil.

Once I understand how a surrogate in a utf-16 encoded file is not being processed properly, that will tell me which error handler is safest to use.  Otherwise unsafe strings of bytes can be fed into Sigil to create epubs that can be used to compromise specific readers which Sigil must protect against at all cost.

Thanks,

Kevin

kevin-zqw:
Hi kevinhendricks

Here is a epub file, you can try generating TOC and see the crash. I assume that you never encountered an issue like this.

Thanks so much.

[Ext.epub.zip](https://github.com/Sigil-Ebook/Sigil/files/2029200/Ext.epub.zip)

kevinhendricks:
Thank you.  That will help prove if the issue really is in the EmbeddedPython interface or not.

kevinhendricks:
Interestingly, I get no crash and have no problems generating a TOC using your Ext.epub.  This is on MacOSX (my main development platform).  Exactly what OS and what version of Python are you using that actually crashes?

BeckyDTP:
I confirm the crash.
Sigil 0.9.9.
Windows 7 Pro, Windows 10 Home, Windows 10 Pro.

kevin-zqw:
same as Becky, sigil 0.9.9 on windows 7, and using the default packed python.

kevinhendricks:
I have attached an untested (never even compiled) version of src/Misc/EmBeddedPython.cpp below.  If either of you get a free moment, please grab, unzip, and rebuild with it and let me know if this was the underlying cause of the crash or not.   If so, I will push it to master when I finally get back to main my development machine. Thanks

[EmbeddedPython.cpp.zip](https://github.com/Sigil-Ebook/Sigil/files/2032201/EmbeddedPython.cpp.zip)

BeckyDTP:
After compile with new EmbeddedPython.cpp file -- no crash.
TOC is created without issues.

kevinhendricks:
Wonderful.  I will push that change tomorrow to master and close this pull request.
Thanks for testing!

Kevin

kevin-zqw:
Thanks so much, @kevinhendricks and @BeckyDTP 

Have a nice day.

kevinhendricks:
Committed EmbeddedPython.cpp to master.  The fix will appear in our next release.  Closing this pull request.  Thank you for your initial bug report and pull request!

BeckyDTP:
I downloaded today latest sources of sigil from github (with last commits) and compiled. It has become a strange thing – a sample file (Ext.epub) again causes a crash when doing the TOC. Please, let somebody check the compilation for Windows, because I tried several times and the problem still occurs, although I already had a file that created the table of contents without any problem. 

I did not change anything in the system, but in the last days there was a cumulative patch for Windows 10. I have no idea what else could happen.

dougmassay:
I'll do some testing. I just can't promise when. Busy weekend/busy month/busy life right now. But I'll do my best! :)

BeckyDTP:
Thank you.
For tests I attach Ext.epub with ncx – this file causes crash already when opening.
[Ext-with-ncx-epub.zip](https://github.com/Sigil-Ebook/Sigil/files/2039955/Ext-with-ncx-epub.zip)

kevinhendricks:
I can confirm that this file causes a Sigil embedded python bug on an unpatched release version of Sigil-0.9.9 on my Mac.  When I get home, I will try it on my build yesterday from master to see if the same issue happens or if it is fixed by the most recent commit.

kevinhendricks:
FWIW,  my build from master with the latest commit has no problems with your new testcase.  My guess is somehow an older version of the EmBeddedPython object file is being linked in to your build somehow.

The new EmbeddedPython.cpp previously attached above fixes any issues I have on my Mac.

BeckyDTP:
@dougmassay If you can build in Windows, please check both files (Ext-with-ncx-epub.zip and Grisham-demo-epub.zip)

@kevinhendricks Thank you for your test. I just need to remind you that issue occurred in Windows. I deleted the EmbeddedPython.cpp.obj file and I successful build. Now issues #356 and #357 are solved, but I see other issue. Please check the attached epub file and look at opf file.

[Grisham-demo-epub.zip](https://github.com/Sigil-Ebook/Sigil/files/2041890/Grisham-demo-epub.zip)

Expected result (clean Sigil 0.9.9 or with all commits but without last):
[HERE](https://i.imgur.com/N8E9Zvy.png)

I get this result (in Sigil with last commit):
[HERE](https://i.imgur.com/3zLh3SG.png)

kevinhendricks:
The last commit is a technically correct fix that can not hurt.  It simply more correctly converts Qt utf-16 QStrings to python3's internal stringstorage that apoears in Python 3.3 and later.  No surrogates should ever appear in correct utf-8 encoded strings. The issue you are seeing is that epub has issues with incorrect xhtml and you have disabled Mend on Open or the Mend code is broken on your build somehow or do to some other latent bug resulting in the opf file not being properly updated.

BeckyDTP:
Mentioned, attached earlier epub file has correct xhtml, although the files are in folders with non-standard names.

Ad 1. Enabled (Open and Save).
Ad 2. Maybe… (I did not touch anything)
Ad 3. Possible.

This morning I tried the last build again and ... everything works. Maybe something else (antivirus?). A strange case. Probably occurs only in exceptional circumstances, but it is not a code issue in Sigil.

I will be watching this build and maybe I will be able to "catch" the problem during normal use.

kevinhendricks:
if you find out anything, please open a new issue.
Closing

kevinhendricks:
BTW, you were right.  After reviewing the last change one more time, I noticed I introduced an unwanted change in how QStringLists are converted to their Python equivalent.

Please pull from master one more time and hopefully all will be well.

BeckyDTP:
I confirm! Thank you!
Ufff… Now I'm calm.
My private build works too.

PR #360, opened 2018-06-05
Leading spaces are ignored in Transifex.

BeckyDTP:
Fix is simple -- add spaces to PluginRunner.cpp.
samples-for-leading-spaces-in-transifex


kevinhendricks:
This looks good and will merge it

PR #372, opened 2018-07-11
Improved translation for statuses in Plugin Runner.

BeckyDTP:


dougmassay:
Aside from the missing tr()'s, these all appear to be _trailing_ spaces in translatable strings, not _leading_ ones. Perhaps trailing spaces are a problem with Transifex, too, but the terminology confuses me.

BeckyDTP:
You're right. _trailing_ (I have to remember.)
Both spaces are a problem in Transifex.

kevinhendricks:
This pull request changes the lgic in an incorrect manner.  Please do NOT mix translation changes with logic changes.  This snippet is incorrect.

0);
@@ -406,6 +406,10 @@ void PluginRunner::pluginFinished(int exitcode, QProcess::ExitStatus exitstatus)
         ui.statusLbl->setText(tr("Status: failed"));
         return;
     }
+    else {
+        ui.statusLbl->setText(tr("Status: success"));
+        return;
+    }	    
 
Therefore this pull request is rejected.

dougmassay:
Fair enough.

Why do you believe the extra else clause you added in PluginRunner::pluginFinished() is necessary?

kevinhendricks:
The final status should not be set until after it passes a number of checks which happens later in that method.  No "return" should be done unless already failed at this point.

BeckyDTP:
The word "success" (as result) is not translated anywhere, this condition seemed appropriate to me.
Of course, I understand that it was not a good idea.

kevinhendricks:
Then replace the final setting of a result string by conditioning on result right at the end.  Your added "return" would prevent all of the needed checks.

Please be much more careful when making PRs.

kevinhendricks:
You could for example replace line 484 with an if condition checking m_result and setting the statusLbl appropriately.  FYI m_result can be "", "failed" or "success" after a plugin runs.

kevinhendricks:
We want to make a release in July yet, and so have to close the tree soon for changes in order to properly rebuild the base.ts so that its changes are reflected in Transifex.  So are you planning on a new version of this pull request and if so how soon?  If so please let us know as we can hold off recreating the base.ts file for a day or so.

BeckyDTP:
I prepared pull request 273.

PR #373, opened 2018-07-12
New version pull request 372

BeckyDTP:
Modified version, without "else".
Word "success" is not translated, but I hope this is understandable for users.

For compatibility reasons, I would also change "Status: Loading" to "Status: loading" (all statuses are lowercase).


kevinhendricks:
Not sure how I closed this, so reopening.

kevinhendricks:
It is a bit confusing as it includes the problem else clause and then removes it, but the final product should be okay.

DiapDealer, should we go ahead and merge this?

dougmassay:
Looks like it checks out to me.

PR #414, opened 2019-04-24
cmake: use GNUInstallDirs

arthurzam:
On Linux systems, use `GNUInstallDirs` for selecting the proper installing location.
The only difference with this PR is that instead of `/usr/lib` it installs to `/usr/lib64` according to the distribution configuration (Gentoo).

Tested on my computer for some days, works without any problems.

Also on another question:
Is it possible to use system `hunspell` dictionaries instead of bundled? On Gentoo and Arch Linux they are located in `/usr/share/hunspell/`.


eli-schwartz:
>  The only difference with this PR is that instead of `/usr/lib` it installs to `/usr/lib64` according to the distribution configuration (Gentoo).

It will also use /usr/lib64 on Arch Linux, which is incorrect as Arch Linux must use /usr/lib

As for the dictionaries, have you read the build documentation? https://github.com/Sigil-Ebook/Sigil/blob/master/docs/Building_on_Linux.md#cmake-options
Arch Linux has built with the system hunspell dictionaries for over a year as per the official Sigil documentation.

dougmassay:
There's too much unknown for me to take a chance on this just now (case in point is Eli's Arch breakage above).

There's also several cmake directives (build-time) and a couple of environment variable (run-time) that alter where Sigil's various pieces will be installed (or can be found). Not to mention some special path "trickery" that allows Sigil's plugin framework to find/use some of the shared libraries that sigil includes (or uses). I've no idea if any of that might be broken with this PR, but I suspect some of it would.

arthurzam:
> As for the dictionaries, have you read the build documentation? https://github.com/Sigil-Ebook/Sigil/blob/master/docs/Building_on_Linux.md#cmake-options
> Arch Linux has built with the system hunspell dictionaries for over a year as per the official Sigil documentation.

I asked just to be sure. Thank you.

> There's also several cmake directives (build-time) and a couple of environment variable (run-time) that alter where Sigil's various pieces will be installed (or can be found). Not to mention some special path "trickery" that allows Sigil's plugin framework to find/use some of the shared libraries that sigil includes (or uses). I've no idea if any of that might be broken with this PR, but I suspect some of it would.

I understand. Maybe we can create a variable to set during cmake configuration. I'm working on cleaning the Gentoo package, and I need to set it to use `lib64`. Maybe something like `LIBDIR`?

dougmassay:
I'll leave this open and look into it. There should be a fairly easy way (at least I _hope_ so!) to accommodate everyone if it's only lib vs lib64 we're talking about, here.

arthurzam:
@dougmassay Big thank you.
I just thought that using `GNUInstallDirs` is the most portable way - guess I was wrong.

> ... only lib vs lib64 ...

Yes, that is mostly what I need.

dougmassay:
> I just thought that using GNUInstallDirs is the most portable way - guess I was wrong.

Not "wrong" _per se_. If Sigil were a brand-new, run-of-the-mill Linux application, GNUInstallDirs (with options for maintainers to easily special-case their install location needs) wouldn't be a problem. But it's a cross-platform monstrosity that several Linux maintainers already have functioning package scripts and patches for. Potentially breaking the several for the needs of the few just doesn't seem wise to me at the moment.

I'll put something together. Probably a default "lib" with a special-cased "lib64" type of solution.

eli-schwartz:
To be fair, with GNUInstallDirs, you can just pass -DCMAKE_INSTALL_LIBDIR=lib and it will go back to installing to `${CMAKE_INSTALL_PREFIX}/lib`. But it's not so simple as to say "it will then always install according to the distribution configuration"... because GNUInstallDirs basically hardcodes some debian stuff and falls back to a cherrypicked lib64, which suits some distros and not others, while the current method of using "lib" will also suit some (different set of) distros and not others.

The advantage to GNUInstallDirs is that it provides a standard cmake configure switch to select which one to use. So it is indeed portable. From my side as the arch sigil packager, this will be a simple change to adapt to.

But I don't know whether, as Doug says, there are any more bits of sigil internals which assume "lib".

a17r:
> > The only difference with this PR is that instead of `/usr/lib` it installs to `/usr/lib64` according to the distribution configuration (Gentoo).
> 
> It will also use /usr/lib64 on Arch Linux, which is incorrect as Arch Linux must use /usr/lib

That is for Arch Linux to consider, as in every other package that is using `GNUInstallDirs`, most likely they don't have to change anything for packaging thanks to this standardisation of path declarations.

dougmassay:
After determining there's nowhere in Sigil's code (or the Python plugin framework) that relies on (or is calculated from) a hard-coded /lib directory, I'm inclined to accept this PR "as is."

The fact that utilizing GnuInstallDirs gives anyone the opportunity to override the locations with standardized cmake configuration switches means easy configuration for everyone. The only problem I see is getting the word out to distro maintainers that there's been a change.

eli-schwartz:
> After determining there's nowhere in Sigil's code (or the Python plugin framework) that relies on (or is calculated from) a hard-coded /lib directory, I'm inclined to accept this PR "as is."

Cool.

> The only problem I see is getting the word out to distro maintainers that there's been a change.

Document it in the building on linux document, and add a changelog entry stating "the build system now uses GNUInstallDirs to determine the best installation libdir on Linux"?

dougmassay:
> Document it in the building on linux document, and add a changelog entry stating "the build system now uses GNUInstallDirs to determine the best installation libdir on Linux"?

Works for me. Any advice on "easily" applying this PR to the development "engine" branch as well? I could probably pull it off if it would work on both branches without any modification, but it will need tweeked a bit.

eli-schwartz:
```
git config --local --add remote.origin.fetch '+refs/pull/*/head:refs/pull/*'
git checkout engine
git fetch
git cherry-pick refs/pull/414
```

 works and applies cleanly for me. Unless there are more bits on that branch which need modification too?

...

On a personal level, I dislike merge requests and do a lot of manual copy-pasting of commit patchsets. Being able to fetch a github PR as a git ref and apply it manually is pretty handy, I find.

dougmassay:
Done (other than updating the documentation). Thanks all.

PR #416, opened 2019-04-29
Add 6 highlight text buttons

Tetralet:
It's not good enough, but it works. Please consider to apply it. Thanks!

2019-04-29-161728_1920x1080_scrot


kevinhendricks:
Why would highlighting text in different colours matter?  What is its purpose?  If the purpose is to change the colour of text, that can and should be done in css.

Also note, that your highlight in colours takes up much of the tool bar area that is actually needed for plugins.  Each plugin can add important functionality to Sigil.

dougmassay:
I agree. The text-formatting buttons that are currently supplied are sufficient.

kevinhendricks:
Closing this as screen realestate  is very tight for space, and the usefullness of the addition is not clear to me.

PR #419, opened 2019-04-30
Dock Validation Results to top too

BeckyDTP:
Enable the option to dock the Validation Results window at the top of the screen.
REQ [here](https://www.mobileread.com/forums/showpost.php?p=3839140&postcount=171)


dougmassay:
I'm not opposed. Just have to make sure it was only for aesthetic reasons that the top dock area was excluded from the list of allowable dock areas for the Validation dockwidget in the first place

PR #426, opened 2019-05-21
Engine

mft39:
I was using Sigil Book Editor for several years. I am wondering if it is possible to make feature to be able to save book at the same time in to different locations. At the same time would be great to have autosave option.


dougmassay:
I'm not exactly sure what you've done here, but this is not a valid Pull Request. These are all old master commits you're trying to merge with the newer engine branch for some strange reason. But to answer your questions before closing it: 

1) Saving to two different locations at the same time is a pretty unique feature to consider making a standard one. I don't really see the point. Do other programs do this? You should be able to make a plugin to do this for you really need this feature, though.

2) Auto-save will likely never be a built-in feature of Sigil. Too many ways for it to totally backfire to be useful.

In the future, please don't submit pull requests that can't possibly be integrated. Use the Issue tracking database for feature requests.

PR #430, opened 2019-06-05
use minizip's ZIP_OK constant instead of zlib's Z_OK

sharkcz:
This is a cleanup making it easier to move to [minizip2](https://github.com/nmoinvaz/minizip)

### DIFF
--- a/src/Exporters/ExportEPUB.cpp
+++ b/src/Exporters/ExportEPUB.cpp
@@ -143,13 +143,13 @@ void ExportEPUB::SaveFolderAsEpubToLocation(const QString &fullfolderpath, const
     fileInfo.tmz_date.tm_year = timeNow.date().year();
 
     // Write the mimetype. This must be uncompressed and the first entry in the archive.
-    if (zipOpenNewFileInZip64(zfile, "mimetype", &fileInfo, NULL, 0, NULL, 0, NULL, Z_NO_COMPRESSION, 0, 0) != Z_OK) {
+    if (zipOpenNewFileInZip64(zfile, "mimetype", &fileInfo, NULL, 0, NULL, 0, NULL, Z_NO_COMPRESSION, 0, 0) != ZIP_OK) {
         zipClose(zfile, NULL);
         QFile::remove(tempFile);
         throw(CannotStoreFile("mimetype"));
     }
 
-    if (zipWriteInFileInZip(zfile, EPUB_MIME_DATA, (unsigned int)strlen(EPUB_MIME_DATA)) != Z_OK) {
+    if (zipWriteInFileInZip(zfile, EPUB_MIME_DATA, (unsigned int)strlen(EPUB_MIME_DATA)) != ZIP_OK) {
         zipCloseFileInZip(zfile);
         zipClose(zfile, NULL);
         QFile::remove(tempFile);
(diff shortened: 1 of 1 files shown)


eli-schwartz:
What is the purpose of switching to this other project instead of zlib? If minizip2 is so good, why not switch to it right now? The change seems to be useless/pointless unless it comes as part of a larger changeset that actually does switch to this other project.

As for actually switching, I've never heard of minizip2 before, but, checking, I see that it's not available in any of the linux distribution repositories I've checked, so I have to ask... what is the purpose of it? What benefit would switching bring to Sigil? The project README suggests the answer is "WinZip encrypted zip files".

dougmassay:
I tend to agree with Eli, here. I don't really see the point. Whether or not minizip2 is accommodated in this one particular regard, zlib/minzip will still be required by Sigil. So unless it's a wholesale change to minizip2 (which won't be happening based on its relative obscurity), there's not very much incentive to accept this piecemeal change.

sharkcz:
First, this change is about consistency, the (original, zlib one) minizip functions return ZIP_ prefixed constants/defines, not Z_*, see eg. at https://github.com/madler/zlib/blob/master/contrib/minizip/zip.c

Second, Fedora is switching to the new minizip (with an active development), see https://bugzilla.redhat.com/show_bug.cgi?id=1632208

kevinhendricks:
I checked the original contributed minizip source and the two versions in our repo and according to the zip.h, the return values should be ZIP prefixed to be technically correct (as sharkcz pointed out).   That said, this change really has nothing to do with Sigil itself moving to minizip2 as epubs do not use passwords, nor WinZipAES encryption, nor splitting, nor any of the "features" mentioned on the minizip fork page.

That also means that this pull request should not have any implications or cause any trouble for our main 3 platforms (Linux, macOS and Windows) and all should continue to build (if Sigil includes the correct header file :)) and work just fine.

So my vote is too merge this pull request as it technically moves to using the properly defined error code return values documented in current minizip's zip.h header file.

Is that okay with you Doug?

If so, we can merge it and cherrypick it to exist in our engine branch as well.

sharkcz:
For the record, the unzipping routines in Sigil already use the minizip's UNZ_* values.

dougmassay:
But will zlib be removed entirely from Fedora? What about the rest of the Z_* constants in Sigil's codebase? Why worry about Z_OK, specifically, when Z_DEFLATED and Z_NO_COMPRESSION are still present? And since zlib is still a Sigil requirement on Linux (when compiling with USE_SYSTEM_LIBS) won't Z_OK still  work just as effectively as ZIP_OK?

I'm not opposed _per se_, I just don't see the problem this tiny change is addressing. Will Sigil not function, as is, with zlib and the new minizip in place (and found by cmake)?

dougmassay:
> Is that okay with you Doug?
It's OK with me. But as I mentioned, should the other Z_* prefixed constants/retvals in Sigil be changed as well?

sharkcz:
zlib naturally stays, only its minizip part will replaced with the fork. Although zlib's (very) slow development is causing some issues. but that's not relevant here.

I've already verified Sigil works fine with the forked minizip, I'm still working on how to best handle the remaining Z_* constant, ideally to avoid explicit inclusion of "zlib.h"

sharkcz:
The fork's maintainer looks interested in providing as good as possible compatibility, so ideally it will become a drop-in replacement.

eli-schwartz:
Changing to fit a documented return code in minizip/zip.h rather than the one coming from the `#include`d zlib.h seems fair enough, but minizip is definitely not compatible as a drop-in replacement -- you would know this since you filed https://github.com/nmoinvaz/minizip/issues/400 due to Sigil using functions in minizip that are not in minizip2!

kevinhendricks:
Currently our minizip includes zlib.h in its zip.h and thus use of the zlib compression flags Z_DEFLATED and Z_NO_COMPRESSION in this code works just fine and are valid.  These flag values are not wrappered by current minizip at all.

The current pull request is valid and more technically correct.  As is the continued use of the zlib compression flags values as zip.h in minizip includes zlib.h.

Any compatibility layer of a minizip2, will need to *not* include zlib.h and will instead need to provide consistent defines for all of the Z_ flags (and not just redefine them as MZ_ flags if it ever wants to be a drop in replacement.

So the pull request as stands is correct, but more changes would be needed for any minizip2 to be considered as the full set of Z_ flags for compression and other things was made available in the original minizip and needed to be used.  That is something for minizip2 to figure out.

sharkcz:
> For minizip2 this could be as easy as including something like the following for each required flag value in its compat headers.
> 
> ```
> #ifndef(Z_DEFLATED)
> #define Z_DEFLATED MZ_DEFLATED
> #endif 
> ```

ack, I'm going to propose that, unless you would be faster :-)

kevinhendricks:
So I am going to merge this pull request as it is technically correct as it stands and has nothing to do with any possible minizip version 2 and should not impact anything.  We can deal with the whole minizip2 idea once the Linux platforms come to some agreement or not at all if minizip2 really does become a drop in replacement that linux packages can use via the current external build process.

PR #434, opened 2019-06-19
Fix to python regex site-package

BeckyDTP:
Setup file does not contain regex for python and plugins with `import regex` do not work.
Look here: https://ci.appveyor.com/project/Sigil-Ebook/sigil/builds/25343465#L1162


eli-schwartz:
Aha, you're probably seeing the update from regex 2019.02.18 to regex 2019.02.19: https://bitbucket.org/mrabarnett/mrab-regex/commits/5d8b8bb2b64b696cdbaa7bdc0dce4b462d311134

+1 on the change.

dougmassay:
Yep. Here's the commit that I made in the webkit branch. 5337db9 

Thought I might have been crazy for a bit. ;)

PR #459, opened 2019-10-04
Proposal for better support for fonts tooltip

BeckyDTP:
rawfont.styleName() is always empty, as a workaround on Windows Italic/Oblique supported by rawfont.style()

For fonts where the family name cannot be read - information about the lack of reliable data (most often for encoded fonts).

"Normal/Regular" is a standard typeface, so this information is redundant.

I tested it on many files with many fonts and it works really well under Windows.
I also added checking that the weight record does not exist in the family name and this has further improved the resulting tooltip.


kevinhendricks:
Please put back the weight-name lookup for Regular/ Normal  as these match what is in qt and it makes no sense to leave it out.  All the rest is fine.  Fix your pr and I will merge it.

kevinhendricks:
looks good, will merge.

PR #482, opened 2019-12-25
Fix source typos

luzpaz:
Found via codespell -q 3 -S ./3rdparty,*.ts,*.dic,*.aff -L doubleclick,pres,te,varn


kevinhendricks:
Thank you

PR #484, opened 2020-01-02
Spell check: Ignore soft hyphens in words

BeckyDTP:


kevinhendricks:
Looks good.   As soon as tests builds pass, I will merge it.

dougmassay:
Any need to include zwj characters while we're at it?

kevinhendricks:
A dictionary that has words that use zwj should have that character its .aff in the wordchars line. Unlike the soft hyphen which is not a valid char in any language.  Are words the use zwj validly spelled without it?

kevinhendricks:
I will merge this and we can add zwj if it turns out to be needed.

dougmassay:
Makes sense. I definitely defer to your expertise on dictionary matters. :)

PR #494, opened 2020-02-07
Remove border in Preview dock window (undocked)

BeckyDTP:
Only today I noticed that after undocking the preview window there is an ugly frame (only on toolbar).
Especially in dark mode, when the white line "glows in the eyes".

sigil-preview-window-remove-border

### DIFF
--- a/src/MainUI/MainWindow.cpp
+++ b/src/MainUI/MainWindow.cpp
@@ -4900,6 +4900,7 @@ void MainWindow::ExtendUI()
 
     m_PreviewWindow = new PreviewWindow(this);
     m_PreviewWindow->setObjectName(PREVIEW_WINDOW_NAME);
+    m_PreviewWindow->setStyleSheet("QDockWidget {border: none;}");
     addDockWidget(Qt::RightDockWidgetArea, m_PreviewWindow);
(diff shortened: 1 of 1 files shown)


BeckyDTP:
I checked it on Windows.
Please check in Linux / Mac.

dougmassay:
Huh. Never really noticed before. I wonder why it's only around the toolbar? All of the dockwidgets except for Preview have a visible border that goes entirely around them (docked and undocked). Perhaps we should see why Preview's only seems to show up around the Toolbar. Maybe instead of changing Preview we should be look at padding on the toolbar or something similar.

BeckyDTP:
I checked _padding_ and _margin_ (no difference, but uglier).
This border applies to the entire preview window and two problems overlap:
a) the webview is without a frame
b) the toolbar has no handle to move

After a few attempts, I decided that removing the border is the most aesthetic solution, since the larger view window has an invisible frame (the webview is from edge to edge of dockwindow).

kevinhendricks:
I do not really notice this at all on my mac, but I am okay with this change.

dougmassay:
I'm not noticing anything on Linux either. I'm going to merge it if it fixes an issue on Windows.

PR #511, opened 2020-03-19
Missing translation in CPCompare

BeckyDTP:
sigil-not-translated


BeckyDTP:
This commit does not require changing _base.ts_, because both expressions ("Current:" and "Checkpoint:") are already contained in it.

PR #514, opened 2020-03-29
Work around dulwich assumption about sys.argv being defined

mihailim:
dulwich.server's main() assumes sys.argv is always defined, however
that's not the case with embedded Python older than 3.8, as discussed
in https://bugs.python.org/issue32573

We can work around that by defining a fallback sys.argv initialized
with an empty list if sys.argv isn't defined before importing any
dulwich modules.

The relevant traceback looks like this:

Traceback (most recent call last):
    File "/usr/share/sigil/python3lib/repomanager.py", line 38, in <module>
        from dulwich import porcelain
    File "/usr/lib/python3/dist-packages/dulwich/porcelain.py", line 121, in <module>
        from dulwich.server import (
    File "/usr/lib/python3/dist-packages/dulwich/server.py", line 1069, in <module>
        def main(argv=sys.argv):
    AttributeError: module 'sys' has no attribute 'argv'

### DIFF
--- a/src/Resource_Files/python3lib/repomanager.py
+++ b/src/Resource_Files/python3lib/repomanager.py
@@ -34,6 +34,11 @@ import filecmp
 from diffstat import diffstat
 from sdifflibparser import DiffCode, DifflibParser
 
+# Work around dulwich assumption about sys.argv being defined,
+# which is not automatically the case with embedded Python < 3.8.
+if not hasattr(sys, 'argv'):
+    sys.argv  = ['']
+
 import dulwich
 from dulwich import porcelain
 from dulwich.repo import Repo
--- a/src/Resource_Files/python3lib/repomanager.py
+++ b/src/Resource_Files/python3lib/repomanager.py
@@ -35,9 +35,10 @@ from diffstat import diffstat
 from sdifflibparser import DiffCode, DifflibParser
 
 # Work around dulwich assumption about sys.argv being defined,
-# which is not automatically the case with embedded Python < 3.8.
-if not hasattr(sys, 'argv'):
-    sys.argv  = ['']
+# which is not automatically the case on Linux with distribution-provided
+# embedded Python versions older than 3.8.
+if (sys.hexversion < 0x03080000) and not hasattr(sys, 'argv'):
+    sys.argv = ['']
 
 import dulwich
 from dulwich import porcelain


kevinhendricks:
I build and test regularly with Python 3.7.2 via the embedded python interface and Sigil and have never seen this error.   Exactly which version of Python are you using and on what platform?

dougmassay:
You're receiving this error in a working version of Sigil? If so which linux flavors and versions of Python? I'm using Python 3.7.2 for all Windows versions of Sigil and don't get that error.

kevinhendricks:
And what exact version of dulwich are you using if this is Linux.

dougmassay:
As stated in the build documentation, it's going to take a very recent version of dulwich to be able to work with Sigil's check-pointing features. The latest (0.19.15) is highly recommended.

kevinhendricks:
We often use that pattern on our main() routines to create commandline standalone test cases.  But main() is never invoked from anyplace but a command line.  Dulwich in server.py is actually using argv=sys.arg as the default in the parameter list unlike our normal code which hides it inside the main routine which prevents it from being invoked during am import.

So I am inclined to include this but probably under either a platform test or python version test, as it is "fixed" in python 3.8 (according to the bug report you cited) but does NOT happen with Python 3.7.2 on either macOS or Windows.

So we really need to know both the python version and platform version you are seeing this bug under, so that I know how to condition the test.

dougmassay:
Makes sense. Is repomanager.py the only place in our embedded routines where we should implement it? I'd hate to run into it again down the road when we use another module that doesn't  protect sys.argv in 'main'.

kevinhendricks:
I checked all python3lib python files and we always protect argv=sys.argv inside of main().  And for plugins, they are always launched in a separate process via the command line so sys.argv will always be available.

I can not check all of the code we install via pip3 but dulwich, css-parser, and lxml are inly invoked by python3lib and of those, since we have never seen the issue before, only dulwich seems tobe the culprit here.

But I do not think dulwich ever assumed it was being run from an embedded version of python.  Obviously, this is really a bug in python but if python devs do not want to fix it in earlier releases, we will have to include the workaround for older Linux platforms.  I just want to know which ones.

mihailim:
Hi!

Running this on Ubuntu 18.04 with Python 3.6.7 (the distro-packaged 3.6.7-1~18.04).

Yup, aware of the new dulwich requirement (currently working on backporting a properly packaged 0.9.15), however dulwich's master still exhibits that behavior, see https://www.dulwich.io/code/dulwich/blob/master/dulwich/server.py#L-1133 , so I thought I'd get ahead of it anyway :)

kevinhendricks:
So please modify your PR to limit it to python versions strictly less than 3.7.2 and we will merge it.

Thanks for the head's up!

mihailim:
Awesome, will do. Thanks!

dougmassay:
I say we do it for Linux with anything less than Python 3.8 to be safe. We control the Python versions on Windows/MacOS.

I can create a 1.2.1 tag (without a full release) that affected linux users/repo managers can use.

dougmassay:
I'm OK with the above suggestion, also.

kevinhendricks:
Good Idea,  please restrict it to all Python Versions less than 3.8.0.

dougmassay:
Sounds good. I'm glad you held off on pushing your reorg for a bit. ;) Will give me a chance to make the 1.2.1 tag.

kevinhendricks:
Looks good to me.  Merging.

PR #519, opened 2020-04-03
Change box size of Preferences color list box

Axa-Ru:
Changed form PAppearanceWidget.ui

PAppearanceWidget ui  master  - Qt Creator_040

### DIFF
--- a/src/Form_Files/PAppearanceWidget.ui
+++ b/src/Form_Files/PAppearanceWidget.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>576</width>
-    <height>475</height>
+    <height>487</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -212,12 +212,28 @@
           </item>
           <item row="1" column="0">
            <layout class="QGridLayout" name="gridLayout_5">
+            <item row="1" column="0">
+             <spacer name="verticalSpacer_3">
+              <property name="orientation">
+               <enum>Qt::Vertical</enum>
+              </property>
+              <property name="sizeType">
+               <enum>QSizePolicy::Minimum</enum>
+              </property>
+              <property name="sizeHint" stdset="0">
+               <size>
+                <width>20</width>
+                <height>20</height>
+               </size>
+              </property>
+             </spacer>
+            </item>
             <item row="0" column="0">
              <widget class="QListWidget" name="codeViewColorsList">
               <property name="maximumSize">
                <size>
-                <width>220</width>
-                <height>160</height>
+                <width>16777215</width>
+                <height>16777215</height>
                </size>
               </property>
              </widget>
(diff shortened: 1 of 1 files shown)


dougmassay:
I'd rather not. The 475 height has proven to ensure that the preference dialog fits the screen on all platforms at most resolutions (regardless of any highdpi scaling). I don't see the need for a blanket 12 pixel increase in the size of one widget.

dougmassay:
Also, why the need for the color list widget to be able to expand indefinitely?

dougmassay:
All this seems to achieve (based on my local testing) is making the color list fill the grid layout horizontally. But all of that increased space is completely empty space. Don't all the attributes already fit with room to spare horizontally in the list widget?

kevinhendricks:
Agreed.  Unless there is clear benefit/reason behind this change (that I am not seeing either), we should probably close this.

dougmassay:
I'll give them a bit of time to make the case for why filling the extra horizontal space (with empty space) would be advantageous, or important, otherwise I'm in agreement. Unless I'm missing something, this will likely be closed.

BeckyDTP:
If I can add something:
* I suggest leaving the height of the entire Preferences window.
* Also, the width of the color window should remain unchanged.
* However, I like stretching the color window and people constantly "tuning" the colors will be satisfied.

Patch:
[patch.txt](https://github.com/Sigil-Ebook/Sigil/files/4428084/patch.txt)

dougmassay:
Your third bullet point makes absolutely no sense to me. Why would people who tune colors (constantly or rarely) care in the least that an ocean of empty space was added to the right of the items in the list? It makes absolutely no difference in the ability to easily tweak the colors.

BeckyDTP:
No, no!
My patch not change width the color list.

The preferences window remembers your size and tab, so if you can see all the colors immediately after pressing F5 (without using the scroll bar in the color list) - life becomes easier.

Of course this is only a cosmetic change, but - I admit - I missed the possibility of extending (vertical) the list of colors.

sigil-preferences-colors-idea

dougmassay:
I'm closing this pull request either way. If someone can create another that maintains the 475 height minimum of the preference widget; doesn't needlessly add an ocean of empty space to the right of list, but allows the color list to expand more in the vertical direction, I'll be happy to include that change (though it seems a little nitpicky to me).

BeckyDTP:
The author of the pull request is probably from Russia.
Look at the interface in Russian and you will understand that his proposal was dictated by the improvement of UI for Russian-speaking users.
I suggest adding a few pixels of the width of the color list to avoid horizontal scrollbar in Russian + vertical expansion.
I can prepare such a pull request.

sigil-preferences-colors-russian

dougmassay:
Yes. I see that the horizontal scrollbar is necessary when Russian is the interface language. Submit your request and I will test it on two platforms. The initial preference dialog (when no user preferences exist) still need to default to maintaining the current height minimums. I can't have the appearances preference widget defaulting to going beyond what will fit in an 800x600 screen resolution for new users.

dougmassay:
Closing in favor of the less intrusive #520

PR #520, opened 2020-04-03
Other try to change box size of Preferences color list box

BeckyDTP:


dougmassay:
This works for me. Thanks.

PR #541, opened 2020-08-10
Fix epub3 mapping functions in bookcontainer and outputcontainer

wrCisco:
Hi Kevin, Doug,
I stumbled upon an error on the epub3 mapping routines in bookcontainer.py and outputcontainer.py. It seems that they remained halfway between invoking the id_to_* dictionaries and the map_id_to_* methods of the wrapper object. I think this commit should fix that.
Thanks!


kevinhendricks:
Yes, nice catch.  I confused the dict name with the access routine name.
Thank you for catching that!

PR #550, opened 2020-09-24
Fixed inspect icon from main theme

BeckyDTP:
Oops! Instead of the "inspect" icon I sent earlier "preview" (from PageEdit).


kevinhendricks:
Thank you!

PR #556, opened 2020-09-27
Nice dotted handle for toolbar...

BeckyDTP:
...in dark mode for Windows

The original one still seems too big to me, so why not change it?
toolbar-handle-dark-mode-windows


dougmassay:
I never really noticed it being "too big" myself, but I'll certainly take a look.

binarom:
I have the impression that the original (consisting of 12 squares) with small sizes of icons overwhelms them. Hence my suggestion for a more delicate, but still visible version of the handle.

BeckyDTP:
I showed some working screenshots to a friend. He wanted four dots, but three should be enough.

dougmassay:
I guess I don't quite understand fully. If the original size "overwhelms the small icons" wouldn't that affect both lightmode and darkmode? The icons are the same size in both light and dark, no?

dougmassay:
It just seems a little too "subject to personal taste" to me. Rather than decide for everone, can't the few people that it bothers just assign their own QToolBar::handle image with qss in their custom stylesheet? I just don't want to get too micro-managy with the defaults, here.

BeckyDTP:
As we use the "Fusion" internal style for dark mode, it is just different from the default mode.

I know this Pull Request is "cosmetic", thus I will leave the decision to you.

dougmassay:
`As we use the "Fusion" internal style for dark mode, it is just different from the default mode.`

Good point. I hadn't thought of that.

dougmassay:
OK. I'm sold. Just for the dark-theme used by Windows (and some Linux users) like you've done it here though. :)

Thanks!

PR #557, opened 2020-09-27
Fix for git-commit icon in main theme

BeckyDTP:
Fix for rasterize

git-commit-rasterize


dougmassay:
Is this just an image tweak? I don't see any weirdness in the old one, but then I don't see a well as I used to, either! :)

BeckyDTP:
Look here:
git-commit-rasterize-big

dougmassay:
I still don't understand. I don't see that horizontal line (or any of those jagged raster edges) in the current git-commit.svg no matter how big I blow it up. But I'll take your word it needs fixing. :)

BeckyDTP:
The icon consisted of two elements - a triangle at the top and a rectangle below. Together they formed an arrow.
When the icons were in PNG files, the problem didn't exist.
When I saved them as SVG, the problem was not visible either, but when Qt rasterizes the icons on the toolbar - both elements building the arrow got an additional border, which I only noticed today.

dougmassay:
I guess I didn't realize Qt was rasterizing them. I assumed it was just downscaling the svg. I certainly can't see any of the artifacts you mention with the naked eye on Linux or Windows at 3840 x 2160.

BeckyDTP:
I can only assume that it is related to a specific situation: operating system + video card + drivers + screen resolution + something else = visible artifact in this icon.
Since I was able to "catch" the problem in the screenshot before, it definitely existed and is now fixed.

PR #558, opened 2020-09-27
Optimize SVG icons

BeckyDTP:
Remove width and height for material icons and optimize SVG files for main icons.


kevinhendricks:
Thank You!

kevinhendricks:
ps What around your material alternative color rcc files?  Do they need to have the height="24" and width="24" removed as well?

PR #582, opened 2021-01-11
Correct splitting lines of text files

BeckyDTP:
It should be more universal, because in Windows all lines are combined into one large paragraph.

### DIFF
--- a/src/Importers/ImportTXT.cpp
+++ b/src/Importers/ImportTXT.cpp
@@ -95,7 +95,7 @@ QSharedPointer<Book> ImportTXT::GetBook(bool extract_metadata)
 QString ImportTXT::LoadSource() const
 {
     QString source = Utility::ReadUnicodeTextFile(m_FullFilePath);
-    source = CreateParagraphs(source.split(QChar('\n')));
+    source = CreateParagraphs(source.split(QChar::ParagraphSeparator));
     return CleanSource::Mend(source, m_EpubVersion);
 }
 


kevinhendricks:
But Utility::ReadUnicodeTextFile is used to load the file and it calls ConvertLineEndings so only \n is left.  What other character are you using to split paragraphs?

kevinhendricks:
Are you copying and pasting text selected in Sigil into a file that will leave in the text  Unicode U+2029 paragraph separator character instead of a newline \n character?

kevinhendricks:
And wouldn't we want both linefeed 0x0a and your QChar::ParagraphSeparator unicode 2029 as split points?

BeckyDTP:
Example.

I don't paste the text file, but I choose Open (* .txt) and open the test.txt file.
[test.txt](https://github.com/Sigil-Ebook/Sigil/files/5798395/test.txt)

Plain text file. Three lines.

If I get everything joined (before.png)  – hard to find where to insert paragraph markers.
before

If I get a view like after.png – IMHO it's easier to add paragraph markers. Of course the after.png view is easy to "destroy" by selecting "Mend and Prettify" because then we'll get everything joined again.
after

---

And so – 0x2029 and 0x000a at the same time are probably a good solution.

It's possible that people open text files in a specific layout, and then the existing solution works better. Personally, I never use this feature, but today I opened a text file and when I saw the result I thought, "It's not for me."

BeckyDTP:
OK. I see.
PR closed without merged.
I checked in the code how paragraph marks are inserted and of course nothing needs to be changed :)

kevinhendricks:
Yes, it splits into lines then is there is an empty line or if the line is indented or has a tab or other whitespace, then a new paragraph is created.

I had to go read that code myself!

PR #593, opened 2021-03-03
Typo: Columbia->Colombia

Tex2002ans:
The South American country is with 2 'o'.

https://en.wikipedia.org/wiki/Colombia


kevinhendricks:
Thank you!

PR #595, opened 2021-03-07
A small change to LinkUpdates

BeckyDTP:
1. Delete unnecessary new lines.
2. Add two indent spaces.


kevinhendricks:
Let's see if we can fix this where it belongs in the GumboInterface routine that performs css updates.

PR #611, opened 2021-04-16
Turkish translation was created

serkan-maker:
sorry i can't work at transsifex.


kevinhendricks:
But we only accept translations from Transifex.  

Translations is not a one time thing.  For every release we will need people dedicated to updating their translations for Sigil.  This is why we use Transifex and not one-off submissions for translations form individuals.  They go away and are not responsive to changes.  Why can't you use Transifex?  It is free.


Sorry but I can not accept this pull request

serkan-maker:
OK thanks

serkan-maker:
OK thanks

PR #621, opened 2021-05-31
doc: fix typo in option name

eli-schwartz:
DISABLE_UPDATE_CHECK is the variable checked for, but the docs and comments exclude the final underscore.

Found this while preparing the latest update.


dougmassay:
Thanks for catching that.

PR #653, opened 2021-11-17
Added checkbox in "Delete Unused Files..." to select/unselect all

luisparravicini:
This is the same change added to the stylesheet selectors dialog but applied to the unused files dialog.


kevinhendricks:
Looks good.  Thank you.

PR #654, opened 2021-11-26
Replace documentation references to the older PCRE lib with refs to PCRE2

mihailim:
Replace references to the older PCRE lib with refs to PCRE2 in the Linux build documentation.

It should be noted that

* Ubuntu 18.04 LTS ships pcre2 10.31
* Ubuntu 20.04 LTS ships pcre2 10.34
* Ubuntu 21.01 ships pcre2 10.36, however since it's going EOL in January 2022 it's not really relevant any longer
* Ubuntu 21.10 ships pcre2 10.37 (goes EOL in July 2022)
* Ubuntu 22.04 LTS (not yet released) is the only one which currently ships pcre2 10.39

### DIFF
--- a/docs/Building_on_Linux.html
+++ b/docs/Building_on_Linux.html
@@ -53,11 +53,11 @@ To get Sigil's Qt5 requirements, <code>sudo apt-get install</code> the following
 <p>Sigil will provide the extra third-party libs if you do nothing, but most (if not all) of Sigil's third-party dependencies should be avialable in your software repos. If you want to make use of them, <code>sudo apt-get install</code> the following packages.</p>
 <ul>
 <li>libhunspell-dev</li>
-<li>libpcre3-dev</li>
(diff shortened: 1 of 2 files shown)


kevinhendricks:
This PR and any timing for it are up to @dougmassay.

There are no releases yet with pcre2 lib and so only current master uses it. As most linux people and all distributions build from official signed releases (ie. 1.8 or etc), the current build docs are correct for those releases.  So updating those docs before Sigil 1.9 is officially released may cause confusion.

mihailim:
Ah, good point about causing needless confusion until post-release! Didn't think about that, went off the recent PCRE2-related commits only. Thank you, closing this then, it's a rather trivial PR anyway and I don't want to cause needless pressure and low level anxiety by leaving something like this open for weeks :)

kevinhendricks:
No problems. And thanks for trying to help!

PR #669, opened 2022-02-06
Correct QComboBox Class signals

JingMatrix:
There are two signals:
activated(int index) and textActivated(const QString &text),
the code made bad use of it.
See https://doc.qt.io/qt-5/qcombobox.html#signals or
https://doc-snapshots.qt.io/qt6-dev/qcombobox.html#signals .

It has caused me bug in my Qt6 build, I was not able to use the include toc function.

### DIFF
--- a/src/Dialogs/HeadingSelector.cpp
+++ b/src/Dialogs/HeadingSelector.cpp
@@ -1024,10 +1024,17 @@ void HeadingSelector::ConnectSignalsToSlots()
     connect(this,               SIGNAL(accepted()),
             this,               SLOT(UpdateHeadingElements())
            );
+    #if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
     connect(ui.cbTOCSetHeadingLevel,
             SIGNAL(activated(const QString &)),
             this,               SLOT(SelectHeadingLevelInclusion(const QString &))
            );
+    #else
+    connect(ui.cbTOCSetHeadingLevel,
+            SIGNAL(textActivated(const QString &)),
+            this,               SLOT(SelectHeadingLevelInclusion(const QString &))
+           );
+    #endif
     connect(ui.left,             SIGNAL(clicked()), this, SLOT(DecreaseHeadingLevel()));
     connect(ui.right,            SIGNAL(clicked()), this, SLOT(IncreaseHeadingLevel()));
     connect(ui.rename,           SIGNAL(clicked()), this, SLOT(Rename()));


kevinhendricks:
We use ship Qt 5.12.x and support back to Qt 5.10 for older Linux distributions, and textActivated does not exist in these versions.  So I cannot accept this pull request as is.

It needs to be conditioned on Qt version as was done in many places in the code.

#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0) 
    
#else
   
#endif

so please fix your pull request and I would be happy to merge it.

dougmassay:
I think the overloaded activated() signal that returned a QString value was dropped in favor of the textActivated() signal in Qt5.14.

So any Sigil code that makes use of the QComboBox::activated(const QString &text) signal will probably need to be found and  ifdefed with:

kevinhendricks:
Actually, both still work up through Qt 5.15

dougmassay:
> Actually, both still work up through Qt 5.15

I'm not seeing an activated(const QString &text) signal in 5.14 or 5.15 in the Qt documentation. Is that just a documentation oversight?

kevinhendricks:
See this code block:

https://github.com/qt/qtbase/blob/5.15.2/src/widgets/widgets/qcombobox.h

```
#if QT_DEPRECATED_SINCE(5, 15)
    QT_DEPRECATED_VERSION_X(5, 15, "Use textActivated() instead")
    void activated(const QString &);
    QT_DEPRECATED_VERSION_X(5, 15, "Use textHighlighted() instead")
    void highlighted(const QString &);
#endif
```

They hide many deprecated interface calls in their docs that still function properly.  We just hide the deprecation warnings until recently.

So activated(const QString&); is valid up to 5.15.X but is no longer valid on 6.X.
So to keep our tree buidling and functioning back to Qt 5.10 we need to if def and we can just ifdef this to < 6.0.0 and it will work just fine on Qt 5.15.X

dougmassay:
Gotcha. Sounds good to me! I'd rather have most of the ifdefs break cleanly before or after qt6 if at all possible anyway. Having a hodge-podge of ifdefs that check against 5.14, 5,15, or 6.0 makes for a mess to have to sort through.

JingMatrix:
Thanks for the discussion, please tell me if it is now a proper compatiable way to fix it.

kevinhendricks:
Looks Great.  Merged in Sigil,  Thank you!

dougmassay:
Looks like there was one more of these in SpellcheckEditor.cpp. But we'd already ifdefed  with:

`#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)`

Do we want to change it to 

`#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)`

for continuity's sake?

Also... what are your thoughts on switching to the new connect syntax moving forward? I don't think we need to try and convert everything all at once, but whenever we need to modify old connects (or create new ones), I think it makes sense to switch to the new format. I think there's a definite benefit in being able to catch errors related to incorrect or missing signals at compile time, rather than waiting for errors to manifest at runtime.

kevinhendricks:
Yes I would ifdef on version 6 just to be consistent.

I am okay with changing things to use then new connect versions as long as they work back to Qt 5.10.  I just never bothered to learn the new connect interface as the old one worked.

kevinhendricks:
I had made this change and commited it.

dougmassay:
Yeah, I never worried about the new format either. I just never really new that the old format doesn't throw compile time errors for non-existent signals. Our conversion to be compatible with Qt6 made me realize how handy knowing at compile time whether or not signals had changed/disappeared would be very handy. Just like this particular issue.

I'm not worried about new for new's sake. I've just always had a bit of trouble following Qt's signal and slot code around in general. Compile-time errors for mistakes I might make will help me in the long run. :)

dougmassay:
And yes: the new syntax was introduced in 5.0.0

https://wiki.qt.io/New_Signal_Slot_Syntax

PR #685, opened 2022-06-04
docs(readme): update winget hyperlink

vedantmgoyal2009:
https://winget.run does not get updated with the winget-pkgs repository and displays older version. hence, switching to https://winstall.app which is updated in (almost) realtime, and provides a better UI/UX too ;)


dougmassay:
Looks good. Thanks! I have something similar in one of the pages on our github hosted website I'll update to match.

PR #699, opened 2022-10-04
fix build with Python 3.11

sharkcz:
This is related to https://bugzilla.redhat.com/show_bug.cgi?id=2083977, which was about 0.9.x, but the issue fixed there is still relevant in the recent code.


kevinhendricks:
If python devs redefined a macro in their headers in a non-backwards compatible point release, perhaps the correct place to file this bug is with the python devs.  Python 3.9.8 and python 3.10 do not have this issue.

kevinhendricks:
From reading the pep on changing macros to functions, I see this change was intentional....  argh!

I guess extra parentheses should not hurt anything for earlier python versions, so I will merge your pull request.

Thank you!

sharkcz:
Thanks, it also brought consistency to sigil's code, because PyUnicode_2BYTE_DATA() and PyUnicode_4BYTE_DATA() couple lines further have already used the additional parens.

PR #737, opened 2023-12-29
Update sigil_zh_CN.ts

HyperPh:
Fix a translation error in sigil_zh_CN.ts


kevinhendricks:
This file is pulled in from Transifex project for Sigil.  So fixing them here is worthless as this file will be updated from Transifex and the fix will be lost.  Contact the appropriate Transifex translation team for your language to get it fixed.

Closing.

PR #742, opened 2024-02-28
Catch AttributeError in addition to OSError in pluginhunspell.py

### LINKED ISSUE #741: [Bug]: Plugins stopped working on Linux with Python 3.12 and system hunspell [solution found]
### Bug Description

On Linux, when using Python 3.12 as the Sigil Python interpreter (24.04 changed to 3.12 as the default recently) while using the system hunspell, all plugins stopped working with backtrace similar to the testplugin one posted below due to a backtrace in the hunspell plugin loader.

It's trivial to reproduce on Linux, just switch the interpreter to 3.12 then run the testplugin (or any other plugin, for that matter), since the affected pluginhunspell.py code path runs for all of them.

Turns out that this is happening because of a `cdll` object behavior change introduced in [this commit](https://github.com/python/cpython/commit/101cfe679fe10b9b662e815999dadf81fcc32c9c) as a response to [this issue](https://github.com/python/cpython/issues/78997): In 3.12.0a4 or newer, a key lookup when treating `cdll` as a dict no longer raises an `OSError`, but instead raises an `AttributeError`.

This was very badly documented in the [Python changelog](https://docs.python.org/3/whatsnew/changelog.html) as "[bpo-34816](https://bugs.python.org/issue?@action=redirect&bpo=34816): hasattr(ctypes.windll, 'nonexistant') now returns False instead of raising [OSError](https://docs.python.org/3/library/exceptions.html#OSError)." in the *Windows* (!!) section of the 3.12.0 alpha 4 release notes -- not even a clairvoyant would figure out it would impact Linux in this manner...

The fix is trivial, also catch `AttributeError` in addition to `OSError` in pluginhunspell.py. Backwards compatible PR submitted at #742 .

### Platform (OS)

Linux

### OS Version / Specifics

Ubuntu 24.04 daily

### What version of Sigil are you using?

2.0.1

### Any backtraces or crash reports

```shell
Traceback (most recent call last):
  File "/usr/lib/python3.12/ctypes/__init__.py", line 450, in __getattr__

    dll = self._dlltype(name)
          ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/ctypes/__init__.py", line 379, in __init__
    self._handle = _dlopen(self._name, mode)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: /usr/lib/x86_64-linux-gnu/sigil/libhunspell.so: cannot open shared object file: No such file or directory

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/share/sigil/plugin_launchers/python/launcher.py", line 317, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/share/sigil/plugin_launchers/python/launcher.py", line 277, in main
    bc = BookContainer(rk)
         ^^^^^^^^^^^^^^^^^
  File "/usr/share/sigil/plugin_launchers/python/bookcontainer.py", line 42, in __init__
    self.hspell = HunspellChecker(wrapper.get_hunspell_path())
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/sigil/plugin_launchers/python/pluginhunspell.py", line 49, in __init__
    self.hunspell = cdll[hunspell_dllpath]
                    ~~~~^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/ctypes/__init__.py", line 457, in __getitem__
    return getattr(self, name)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/ctypes/__init__.py", line 452, in __getattr__
    raise AttributeError(name)
AttributeError: 
/usr/lib/x86_64-linux-gnu/sigil/libhunspell.so
```

mihailim:
Catch AttributeError in addition to OSError in pluginhunspell.py to work around Python 3.12.0a4 ctypes.cdll behavior change introduced in [this commit](https://github.com/python/cpython/commit/101cfe679fe10b9b662e815999dadf81fcc32c9c) as a response to [this issue](https://github.com/python/cpython/issues/78997): In 3.12.0a4 or newer, a key lookup when treating `cdll` as a dict no longer raises an `OSError`, but instead raises an `AttributeError`.

Fixes #741 .

This change is backwards compatible with any Python version.

Alternatively, instead of `cdll[hunspell_dllpath]` we could use `cdll.LoadLibrary(hunspell_dllpath)` there, but:

* I'm not too sure of the implications for older Python versions and for Windows, since having fspath-like objects work there is a recent ctypes addition
* I suspect that may introduce trouble on Windows -- for Linux we'd need the full path to the .so, while on Windows we must not use that

so let's be cautious and just catch the new exception, since this way we won't need to rely on any dynloader implementation subtleties.

### DIFF
--- a/src/Resource_Files/plugin_launchers/python/pluginhunspell.py
+++ b/src/Resource_Files/plugin_launchers/python/pluginhunspell.py
@@ -47,7 +47,7 @@ class HunspellChecker(object):
         try:
             # First use bundled hunspell location.
             self.hunspell = cdll[hunspell_dllpath]
-        except OSError:
+        except (OSError, AttributeError):
             # No bundled (or incompatible bundled) libhunspell found.
             # found. So search for system libhunspell.
             self.hunspell = None


kevinhendricks:
Has this python3 change appeared in any production release of Python yet (non-alpha, non-beta)?  Why would they make a backwards incompatible change in a point release without any deprecation warnings?  What python bug was that commit meant to actually fix?

Unfortunately your pull request is incomplete.  Your change is needed in other places in that file, and in the ml (multiple languages) version of the hunspell plugin in multiple places as well sigilgumbo library and possibly others.

So please revise your pull request to at least make all the changes where needed in the file you targeted, and I will merge it, and track down all of the other places where cdll is used.

Thank you for your bug report and PR!

The python3 crowd should stop making backwards incompatible changes in minor point releases.  It is a very bad practice.

mihailim:
Oh, as it turns out, the patch for this file at least *is* complete. The other `cdll` reference, i.e. `self.hunspell = cdll.LoadLibrary(sys_hunspell_location)`, doesn't suffer from the same problem: the behavior when trying to dynload a non-existent library via `LoadLibrary()` hasn't changed, still throws `OSError` so it's handled correctly. They only changed the behavior when trying to access an *attribute* of `cdll`, to make it behave like a "true" dict would behave (i.e. they only changed the `__getattr__` behavior.) To be fair, that's more uniform indeed, but this should absolutely have been documented.

`pluginhunspellml.py` is the only other place where we use `cdll` as a dict, also just on line 64 with `self.hunspell = cdll[hunspell_dllpath]`.  `sigil_gumboc.py` does the LoadLibrary() dance so it's not affected.

kevinhendricks:
Great detective work.  Thank you.

dougmassay:
As an Arch user, I'm amazed (and a little thankful) that they've shown as
much restraint as they have with Python 3.12. They're taking their time
(which is usually against their nature) to make sure that packages that
will be affected (the removal of the imp module in 3.12 being another major
point) are updated first.

But it looks like they're going to move on it fairly soon.

***@***.***/thread/J56DVVEVTSPPV3LNXVEQ6AEZUQFZLHMI/

mihailim:
BTW \#2, still for our peace of mind and taking advantage of the opportunity: I've messed around with my usual workflows and have not come across any other Sigil problem with Python 3.12, so I think we should be free from major, uh, unexpected improvements :) Of course, individual plugins may or may not be affected, but the Sigil core plugin infra looks good for 3.12. Small mercies.

mihailim:
Oh, and I'll take care of alerting the Debian Sigil maintainer of the issue and maybe trying to upstream 2.1 once it's out.

kevinhendricks:
Thank you.  Feel free to point to your commits for cherry picking into Debian 2.0.2 as 2.1.0 will not be officially out until just after the end of March.

PR #749, opened 2024-04-05
toolbar extension better icon

BeckyDTP:
The icon at the end of the toolbar in dark mode in Windows is virtually invisible, as its default color is almost identical to the background color of the toolbar.
image

The change is to enlarge the icon and to make it more sharply colored, also when the mouse is hovered over.

I previously mentioned it [here](https://www.mobileread.com/forums/showpost.php?p=4409290&postcount=67).


kevinhendricks:
Looks good!  Thank you!

dougmassay:
I tried to merge this earlier when I was out and about, but my phone was logged out of github. Thanks!

PR #767, opened 2024-08-13
css style applied for sigil_user_guide title page

alsed:
It has always bothered me why, having such beautiful fonts available, the title page of the user guide looks so rough. So, instead of complaining, I decided to do something about it:
* header style used for User's Guide title page
* css style included for the TOC title in it's page.

Hope you like it.


dougmassay:
3 Things.

1) The Sigil User Guide has its [own repository](https://github.com/Sigil-Ebook/sigil-user-guide)

2) The README in the user guide indicates that we're not interested in purely artistic/aesthetic contributions. We're after content and corrections

3) We can't see what your pull request consists of without downloading the epub you created and doing an eyeball comparison with the old one. It sort of defeats the point of using github to compare.

Feel free to submit a pull request in the user guide's repository using the source content there and we'll consider it. But there's a good chance we won't if it's just personal aesthetic preference.

dougmassay:
It has nothing to do with liking it or not. I didn't look at it. Because I couldn't do an online comparison using GitHub. If you submit your changes via the user guide repository, I will be happy to look at it. There's just no guarantee that they will be incorporated. As mentioned, we're trying to avoid contributions that are purely aesthetic in nature. Not because we're in love with the look we have now, but because personal preferences, fonts, and visual appeal are purely subjective, and we don't want to spend our limited time dealing with changes to "the look."

alsed:
I appreciate you have taking the time to explain the rationale behind this decision. 
As requested, the proposed changes had been pushed in the correct guide repository.

Just my 2 cents; I agree that there is now way to reach consensus about "look and feel" design choices. It's better to have people who know about the matter (and I'm also not one), make the final decisions.

PR #772, opened 2024-09-04
Fix when cancelling external editor selection

BeckyDTP:
Bug reported [here](https://www.mobileread.com/forums/showpost.php?p=4450987&postcount=1).


kevinhendricks:
Thank you!

PR #778, opened 2024-10-19
Remove implicit QChar conversions

TheComputerGuy96:


kevinhendricks:
Okay, that makes sense for Arch and Manjaro and Qt 6.8

PR #779, opened 2024-10-19
Replace deprecated setTimeSpec() calls

TheComputerGuy96:


kevinhendricks:
Unfortunately, the correct solution for the "unused variable" is to properly use it at the end of that routine.  That said, the GetInfo routine is a new addition yesterday and is testable but not complete.

I have pushed the proper solution to master.

Please remove your incorrect unused variable "fix", and I will go ahead and merge the remaining part of your PR.

kevinhendricks:
Actually since Sigil supports as far back as Qt 6.4.2 and QTimeZone did not finish that class until Qt 6.5, and because setTimeSpec will not even be deprecated until 6.9,  I am not sure we can merge your PR at all.  It would need to be tested on Qt 6.4 to see if it works properly.

dougmassay:
I approved the Linux CI build of this PR, so we should at least see if it will build with Qt6.4

kevinhendricks:
So it appears to work with Qt 6.4.
So we can merge this once amended to remove the improper fix for the "unused variable".

dougmassay:
It builds ok, but I'm not sure that's enough. It wouldn't be the first time I saw a Qt problem that didn't manifest until runtime. ;) How would I test that this works with a Qt6.4 based Sigil?

That said: we can probably merge the PR for now. If we run into problems later, we can deal with it.

kevinhendricks:
But we only want 7eba71a

We do not want 057540b, as that variable is no longer unsed.

dougmassay:
Got it. So we're still waiting for PR to be changed to reflect that.

kevinhendricks:
Yes, unless you know how to merge only a part of a PR!

dougmassay:
I might be able to cherry-pick the one commit if they don't get back to us.

kevinhendricks:
You can test it by editing any epub and saving it and then check the metadata for modification time to see if it fits the required format and seems reasonable.

kevinhendricks:
Let's give them a chance to fix the PR, otherwise we will cherry pick the part we want.

dougmassay:
Looks like the PR might be incompatible with the current HEAD. Is that just because of the commit in question?

kevinhendricks:
Yes, it removed an "unused" variable in the newly created GetInfo() but instead of removing it, it should have been used to restore BoikBrower state.

TheComputerGuy96:
I was actually sleeping during this time (anyway I removed the unused variable change)

kevinhendricks:
Thank you!

PR #803, opened 2025-04-12
CleanFileName function + improved SaveAs

BeckyDTP:
The title of the ebook from the metadata is the proposed file name when saving instead of the standard untitled.epub. In the proposed file name, characters forbidden by the operating system and spaces are replaced by an underscore character.


dougmassay:
Never really saw the need for Save As to be pre-populated with anything more meaningful myself. I find 'untitled' to be just a tad safer. But I'm not opposed in principle. :)

My only concern would be the (not terribly unlikely) possibility that the metadata produces the same file name as the epub that was originally opened. Under those circumstances, Save As would be trying to overwrite the original. There would be a warning in that case, of course, but still... something to think about.

BeckyDTP:
Here exactly nothing changes. If you open any file and it already has a name it is proposed.

The improvement is practically useful for people who (although I know that you do not recommend it) create files from scratch.
And if they set a title for themselves in the metadata, then when saving they have the pleasant surprise of being offered a file name based on the publication title instead of the neutral "untitled.epub".

So we can say that SaveAs has all the previous functionality + proposing a name if only there is a title and the file has never been saved before.

dougmassay:
Then apparently I have NO objections. :)

dougmassay:
For the record: I have no problem in the world with folks creating Epubs from scratch using Sigil. I just find it odd that anyone would use Sigil to write a book. ;)

BeckyDTP:
Commit 'niewazne' is unimportant. Just git sometimes drives me crazy.

kevinhendricks:
Looks good to me too.  Just having the arbitrary string to filename is a good utility routine to have!

Thank you!

PR #816, opened 2025-06-16
Bug in Save Automate Lists...

BeckyDTP:
..when the automateXX.txt file does not exist.

Bug: Crash when saving empty automation lists for non-existent files.

Fix: Added empty data check in saveData, safe handling in SetNewAutomateList, and exception handling for CannotOpenFile.

Note: Save by Utility::WriteUnicodeTextFile moved to saveData (probably more logical).

### DIFF
--- a/src/Dialogs/AutomateEditor.cpp
+++ b/src/Dialogs/AutomateEditor.cpp
@@ -37,6 +37,8 @@
 #include "Misc/SettingsStore.h"
 #include "Misc/Plugin.h"
 #include "Misc/PluginDB.h"
+#include "Misc/Utility.h"
+#include "sigil_exception.h"
 #include "Dialogs/AutomateEditor.h"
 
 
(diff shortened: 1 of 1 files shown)


BeckyDTP:
Indeed. At first I wanted to make it even simpler: if the data is empty then just close the window. 
But then I thought about error handling and probably overdid it. Since the data is empty and you know that the file would be empty too.

kevinhendricks:
How about we keep your QMessageBox warning piece idea.  It can not hurt.

So putting something like the following inside that if before the accept() and return.

QMessageBox::warning(this, tr("Warning"), tr("Cannot save an empty automation list."));

kevinhendricks:
If you want, I can merge this PR in and manually add back the warning QMessage bit.  Or I can wait and merge it afterwards.

Your choice.  Just let me know which you prefer.

Thanks

BeckyDTP:
Sure. It's just that it's now 3 a.m. in Poland :)

kevinhendricks:
Thank you!

PR #821, opened 2025-08-13
Missing book_lang from OPF for AriaClips

BeckyDTP:
I noticed this today while playing with footnotes in AriaClips.


kevinhendricks:
Nice catch!  Thank you!

PR #825, opened 2025-08-22
Fix displaying ampersand in QTabWidget tab titles – double '&' in tab…

BeckyDTP:
… text

This patch fixes the display of ampersands ('&') in tab titles within `QTabWidget` in Sigil. According to Qt documentation, a single '&' in a tab label is interpreted as a mnemonic (keyboard shortcut), where the following character becomes the shortcut key (e.g., "Bro&wse" → Alt+W). To display a literal ampersand, it must be escaped by doubling it ("&&").

Details: Sanitize tab label with `.replace("&", "&&")`.


kevinhendricks:
Thank you!

PR #826, opened 2025-08-27
Fix for proposed filename...

BeckyDTP:
...when we use templates.

I hope that is what [this issue](https://www.mobileread.com/forums/showpost.php?p=4533343&postcount=1) is about.

### DIFF
--- a/src/MainUI/MainWindow.cpp
+++ b/src/MainUI/MainWindow.cpp
@@ -2058,7 +2058,7 @@ bool MainWindow::SaveAs()
     } else {
         proposed_name = QFileInfo(m_CurrentFilePath).fileName();
         if (proposed_name.isEmpty()) {
-            proposed_name = DEFAULT_FILENAME;
+            proposed_name = (m_CurrentFileName.isEmpty())?DEFAULT_FILENAME:m_CurrentFileName;
         }
     }
 


kevinhendricks:
Wouldn't we still want to use a title based name when m_CurrentFileName is empty as well?

So above where you just changed, should it be ? ...


```
if ((m_CurrentFileName.isEmpty() || m_CurrentFileName == DEFAULT_FILENAME) && !sanitized_title.isEmpty() && sanitized_title[0] != '[') {
        proposed_name = sanitized_title + ".epub";
    }  else {
        ... 
     }
```

Alternatively, I could try to change FolderIn to use your proposed name approach of sanitizing the title or I could change the Template plugins to always use Template2_ and Template3_ as titles.

Thoughts?

BeckyDTP:
As for the extended condition, I'm in favor of it. 👍

> Alternatively, I could try to change FolderIn to use your proposed name approach of sanitizing the title or I could change the Template plugins to always use Template2_ and Template3_ as titles.

The questioner is correct – if the name appears in the title bar (as is the case with the Template2, Template3, and FolderIn plugins, or after opening an existing EPUB file), let's use that name instead of the title in the metadata. We do the same thing with any EPUB file that is opened – if it already has a name, we won't impose a new one taken from the metadata.
In addition, the questioner clearly suggests that they are using a modified plugin (probably Template2), and we cannot control what other plugins do.

So let's assume that my original assumption was correct:
- the user created a new file and did not provide a title in the metadata: untitled
- the user created a new file and provided a title in the metadata: title from metadata
- the user opened an existing file or used the Template2 or Template3 plugin: existing file name
- the user used the FolderIn plugin: name of the folder where the files were located

BeckyDTP:
OK. I understand. You can add a condition to the FolderIn plugin that if the parent folder is called “src,” use the title from the metadata. This change + PR should cover all cases (I hope).

kevinhendricks:
Okay can you please modify your PR to properly include the extended condition and I will merge it.  I am away from my computer so I can not modify your PR myself.  Thanks!

kevinhendricks:
Please add parentheses to make the order clearer since being empty is not enough alone to trigger using the title.

So 
(empty or default) and ...

BeckyDTP:
Sure.

kevinhendricks:
Perfect!  Thank you!
Merging

dougmassay:
I've never really understood the reliance upon having what amounts to a "Save As" come prepopulated with relevant filenames already present. I can't be much help here, I'm afraid. I'd just make every Save As dialog start with "Rename Me" and move on. 

But that's just me. :)

Issue #833, opened 2025-11-11
[Bug]: Sigil on Windows won’t insert some Polish characters (ł, ą, ć, ń) in editor/Find, and regex treats them as non-letters


lamedqoof:
### Bug Description

On Windows with the Polish (Programmers) layout I see two possibly related problems:

1. In the main editor and in the Find field I can type some Polish characters (ę, ó, ś, ź, ż), but other Polish characters (ł, ą, ć, ń) do not appear at all — the keystroke is ignored.
2. In regex searches, Polish letters are not treated as “word characters” by `\w`, so patterns that should match Polish words don’t work, but the same pattern with Unicode properties does.

This makes it hard to edit and search Polish EPUBs.

**Environment**
- Windows 10/11
- Keyboard: Polish (Programmers), AltGr for diacritics
- Sigil: 2.6.2

### 1) Input problem (editor + Find)

**Steps to reproduce**
1. Open any EPUB.
2. Click in the editor and type (with Polish – Programmers):
    - `AltGr+e` → `ę` → inserted
    - `AltGr+o `→ `ó` → inserted
    - `AltGr+s` → `ś` → inserted
    - `AltGr+x` → `ź` → inserted
    - `AltGr+z` → `ż` → inserted
    - `AltGr+l` → `ł` → **nothing**
    - `AltGr+a` → `ą` → **nothing**
    - `AltGr+c `→ `ć` → **Clip Editor opens**
    - `AltGr+n` → `ń` → **nothing**
3. Open Find/Replace, put the cursor in Find, repeat.

**Actual result**
Some Polish characters are accepted, others (`ł`, `ą`, `ń`) are silently ignored, and `ć` is taken by the Clip Editor shortcut.

**Expected result**
All characters produced by the active Windows layout should be accepted in the editor and in all text fields (Find/Replace, etc.), and AltGr-based characters should not be overridden by default shortcuts.

### 2) Regex problem (Polish letters not in \w)

**Example text:**
`odejście`

**Regex that fails:**
`odej\wcie`
→ finds nothing (so `ś` is not seen as part of `\w`)

**Regex that works:**
`odej\p{Ll}cie`
→ matches correctly

This suggests the regex engine in Sigil supports Unicode properties (`\p{L}`, `\p{Ll}`) but the default word class `\w` is ASCII-only, which is inconvenient for Central European languages.

**Expected behavior**
[...]

PR #834, opened 2025-11-12
Fix for #833

BeckyDTP:
In rare cases, the “Unicode Property” option is not visible in the Find & Replace window.


kevinhendricks:
Thank you!

PR #838, opened 2025-12-11
Fix for dots in the middle of file names

BeckyDTP:
Improved file name cleaning, especially for dots in the middle of the name.
Additional replacements for commas and semicolons.

Test name: `Batman vs. Superman`
Now: `_`
Should be: `Batman_vs_Superman`

### DIFF
--- a/src/Misc/Utility.cpp
+++ b/src/Misc/Utility.cpp
@@ -1569,29 +1569,25 @@ QString Utility::CleanFileName(const QString &name)
     result.replace("\"", "_");
     result.replace("+", "_");
     result.replace(" ", "_");
+    result.replace(",", "_");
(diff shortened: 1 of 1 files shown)


kevinhendricks:
FWIW, on Linux there are many valid file names with multiple extensions such as junk.tar.gz, junk.tar.xz, etc.

Most titles will not have multiple "."s in them, will they?   But they are legal in file names and not just one before the extension.

BeckyDTP:
This is not a problem.
This function is universal and does NOT apply to the extension(s).
We only add “.epub” itself when saving the file in `MainWindow::SaveAs()`.

kevinhendricks:
Thank you.

PR #845, opened 2026-02-08
Fix undefined name error

naglis:


BeckyDTP:
I completely forgot about the existence of the hrefutils.py file.
The `getRelativeHREF` function is only used by the _RemoveInLineStyles_ plugin with the `None` parameter for `afragment`, which is why I never found this bug.

You can also add an additional sample test under `main()`:

```
    p1 = 'OEBPS/Text/book1/chapter 1.xhtml'
    f1 = 'page015'
    print('Testing getRelativeHREF(apath, afragment)')
    print('    apath:     ', p1)
    print('    afragment: ', f1)
    print(getRelativeHREF(p1, f1))
    print('    ')
```

kevinhendricks:
Yes, this fix is good.

Merging it.

Thank you.

kevinhendricks:
FWIW, adding a self test in main would be nice as BeckyDTP suggested.  Happy to merge a separate PR for that as well.

dougmassay:
I cancelled the Windows ci run on these two PRs by the way. I've got a Windows build queued up on a beta runner (vs2026) and didn't want all the Windows stuff to stack up on that beta runner.  Just in case anyone wonders. I may cancel others as well. Or I may temporarily disable to automatic Windows ci runs.

kevinhendricks:
Understood.  Thanks

naglis:
> You can also add an additional sample test under `main()`:
> 
> ```
>     p1 = 'OEBPS/Text/book1/chapter 1.xhtml'
>     f1 = 'page015'
>     print('Testing getRelativeHREF(apath, afragment)')
>     print('    apath:     ', p1)
>     print('    afragment: ', f1)
>     print(getRelativeHREF(p1, f1))
>     print('    ')
> ```

Are these run during CI? Perhaps it would be better to add tests utilizing `unittest`?

kevinhendricks:
No they are run in that module's main() (which is never used when accessing it as a module) so that the standalone python file can be run and functional tests are done with no need for Sigil or a plugin environment or unit test.   Simple fast and standalone.

naglis:
Would you be open to moving them to unit tests or at least running them during CI, so that problems would be discovered automatically?

kevinhendricks:
You are assuming it changes often, and that it is only used in Sigil.  When in fact hrefutils.py has been used in a number of my projects over the last 5 to 10 years or so and rarely changes and it has never had a true regression, just untested code paths (that are rarely used as its primary role is in opf parsing to extract manifest item hrefs which can not contain a fragment at all).  

So keeping it and all testing pretty much self contained in that module is important to me and easily achieved.  I will just push Becky's additional functionality test myself.  No worries.

FWIW, our internal plugin support code has been reasonably stable while maintaining backwards compatibility.  Dulwich just moved an entirely public routine (reset_index) with no thought to backwards compatibility in a point release. He always unit tests everything but still broke things horribly for everyone using his software.  We have to keep our plugin internal support code stable across multiple versions of multiple python modules and update and maintain a wrapper version so plugins can tell if a bug fix or a change is included.  So far we have survived.  Bleeding edge distros like Arch reveal most of the issues long before they impact our much larger Windows and Mac user bases which use our embedded python.

PR #846, opened 2026-02-08
Remove duplicated dict key

naglis:
`application/x-font-ttf` key is duplicated.


kevinhendricks:
Yes, duplicated key is a mistake.
Patch looks good.  Merging it.
Thank you.

PR #857, opened 2026-03-30
Refreshing plugin icons on toolbars after making changes to... 

BeckyDTP:
plugin assignments in Preferences

From the Sigil-modified-version

### DIFF
--- a/src/Dialogs/PreferenceWidgets/PluginWidget.cpp
+++ b/src/Dialogs/PreferenceWidgets/PluginWidget.cpp
@@ -95,6 +95,10 @@ PreferencesWidget::ResultActions PluginWidget::saveSettings()
     }
     settings.setPluginMap(pluginmap);
 
+    if (pluginmap != m_OriPluginMap) {
+        results = results | PreferencesWidget::ResultAction_RefreshToolBarPlugins;
+    }
+
     if (!bundledInterpReady()) {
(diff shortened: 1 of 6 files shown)


kevinhendricks:
FWIW, there are two routines in MainWindow that invoke Sigil Preferences.  One goes directly to ManagePlugins and the other is for the user launching Preferences directly.

Search for "prefers" in MainWindow.cpp.  The two routines are adjacent in the code.  The ManagePlugins prefs upon return unconditionally runs LoadPlugins at the end.

We could easily do the same for the other one.  The loadPlugins code is fast. So we do not want to add a bunch of supporting changes in Preferences just to decide whether to run a fast loadPlugins or not routine.  We just change the call to UpdateTooltipPlugins to loadPlugins and remove the unneeded  changes in Preferences.

Just keeps the code simpler.

kevinhendricks:
Another option is to go with this change (adding those changes to Preferences) but then also making parallel changes for the other return from Preferences.

Either way would work but I want to keep the two prefers objects used in MainWindow completely in parallel.

kevinhendricks:
Looks good.  Merging it!  Thank you!  Based on the translated ReadMe  I think there is supposed to be code someplace in sigil-modified to auto assign plugins to avail slots in the quicklaunch plugin toolbuttons if you are interested.  Probably someplace in Manage Plugins Preferences.

PR #862, opened 2026-04-24
Zoom To Fit without triggering scrollbars

BeckyDTP:
Windows 10.
I’ve noticed that after using ‘Zoom To Fit’, the image is slightly too tall (or wide), and causes a scrollbar to appear.
Calculating the space using the viewport solves the problem (for me).


kevinhendricks:
Thank you!

PR #866, opened 2026-05-22
Add JSON file editing and syntax highlighting

drmhessler:
## Summary

Adds basic JSON support to Sigil:

- recognizes `.json` as `application/json`
- maps `application/json` to `MiscTextResource`
- opens JSON files in Sigil's text editor
- adds JSON syntax highlighting for keys, strings, numbers, literals, and structure characters

## Testing

- `git diff --check`
- `cmake --build build --target sigil -j2`

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -377,6 +377,8 @@ set( MISC_FILES
     Misc/XHTMLHighlighter2.h
     Misc/CSSHighlighter.cpp
     Misc/CSSHighlighter.h
+    Misc/JSONHighlighter.cpp
+    Misc/JSONHighlighter.h
     Misc/PythonHighlighter.cpp
     Misc/PythonHighlighter.h
     Misc/HTMLEncodingResolver.cpp
--- a/src/Misc/JSONHighlighter.cpp
+++ b/src/Misc/JSONHighlighter.cpp
@@ -1,0 +1,150 @@
+/************************************************************************
+**
+**  Copyright (C) 2026  Sigil Contributors
+**
+**  This file is part of Sigil.
+**
+**  Sigil is free software: you can redistribute it and/or modify
+**  it under the terms of the GNU General Public License as published by
+**  the Free Software Foundation, either version 3 of the License, or
+**  (at your option) any later version.
+**
+*************************************************************************/
+
+#include <QApplication>
+
+#include "Misc/JSONHighlighter.h"
+#include "Misc/Utility.h"
+
+enum JSONBlockState {
+    JSONNormalState = 0,
+    JSONStringState = 1
+};
+
+JSONHighlighter::JSONHighlighter(QObject *parent)
+    : QSyntaxHighlighter(parent)
+{
+    SettingsStore settings;
+    if (Utility::IsDarkMode()) {
+        m_codeViewAppearance = settings.codeViewDarkAppearance();
+    } else {
+        m_codeViewAppearance = settings.codeViewAppearance();
+    }
+}
+
+void JSONHighlighter::do_rehighlight()
+{
+    QApplication::setOverrideCursor(Qt::WaitCursor);
+    rehighlight();
+    QApplication::restoreOverrideCursor();
+}
+
+void JSONHighlighter::highlightBlock(const QString &text)
+{
+    bool in_string = previousBlockState() == JSONStringState;
+    int string_start = in_string ? 0 : -1;
+    bool escaped = false;
+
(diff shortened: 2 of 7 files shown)


dougmassay:
Can I ask how adding json files to epubs (and editing them) would be useful?

That's an honest question--I'm not trying to be a smart-ass.

kevinhendricks:
Thank you for your PR.  Please point me to where in the epub 3.3 spec json is a supported media-type, or how this is useful for supporting epub3.

kevinhendricks:
Based on a recent web search:

No, raw JSON (application/json) is not an official or "Core Media Type" in the [EPUB 3.3](https://www.w3.org/TR/epub-33/) specification. While you can use JSON for data blocks inside scripts in XHTML, standalone .json files are considered "foreign resources" and require strict fallbacks.

How JSON is Handled in EPUB [3](https://www.w3.org/TR/2021/WD-epub-33-20210917/).3:

Core Media Types: The official core types are strictly limited to XHTML content documents, SVG, CSS, JavaScript (application/javascript), and specific image/audio/video formats (like JPEG, PNG, MP3).Foreign Resources: Any unlisted file type, including .json, falls into this category. If you include a JSON file in your manifest, the specification requires you to provide a fallback attribute pointing to an official content type (such as an HTML alternative).

JSON embedded in HTML: You can embed JSON-LD metadata or data structures, but this must be done directly inside standard <script> tags within an XHTML document, rather than linking to an external .json file.

Recommended Practice: Because reading systems do not widely support or implement foreign resource fallbacks for arbitrary text formats, including a .json file directly in the publication is discouraged. Instead, it is highly recommended to embed your JSON inline within your HTML documents.

kevinhendricks:
FWIW, if we do decide to use this PR, the code changes look just fine (nicely done!).

We just need a good reason to add it since including a .json file anyplace outside of the META-INF folder will not be used or allowed by most e-readers.

Update: just checked the epub 3.4 spec just published on May 6 and json files are still **not** considered a core media type that can be safely used inside an epub (outside the META-INF folder) and have any chance of passing epubcheck.

Still if this is common practice in commercial epubs or supported inside some mainstream epub3 e-readers, we could still add this.

Just need to hear about how it is useful in some way.

drmhessler:
While JSON is not defined as a core EPUB publication resource media type, EPUB3 publications increasingly include auxiliary resources used by scripted content and interactive workflows.
The intent of this PR is therefore not to introduce JSON as a new core EPUB type, but simply to allow Sigil to inspect and edit packaged JSON resources that are already present in modern EPUB3 publications.
This is especially useful for scripted EPUB workflows involving synchronized media, interactive overlays, search metadata, client-side state, or other JavaScript-driven functionality.

dougmassay:
Where do these json files typically reside in EPUB3s? I guess I'm not
opposed to incorporating the functionality (especially if people are going
to start encountering this sort of thing in the wild). But Sigil will
definitely have issues if they're in the meta-inf folder. If not there...

I've also seen some discussion where the recommendation is to manifest
these files in the usual oebps or text folders, but not to include them in
the spine. Not sure if that would require any special-casing within Sigil.

Getting our hands on a commercial epub that uses json in the ways you
describe might need to be the first step. Do you have any examples?

kevinhendricks:
I understand the usefulness of json, but epubcheck will complain about its use anyplace outside the META-INF folder and most commercial publishers still require passing epubcheck before submitting an epub to be published.

Let me think about it, as I have never seen an actual commercial epub3 with a separate json file.  If you could cite an examples or point me to an actual commercial epub that contains one, that would help.

kevinhendricks:
It seems the epub 3.4 spec and epubcheck might allow a json file to be included in an xhtml script tag as a src.

See here:

https://github.com/w3c/epubcheck/issues/1654

and it not require a manifest fallback!

So even if it is not accepted yet, it is coming from the look of things.

kevinhendricks:
Okay based on that epubcheck change for epub 3.4, I will merge this PR.

That said, I would still love a link to a commercial epub that uses a json file for javascript code that is not just a link to an application/ld+json additional opf metadata that is typically a remote link.

drmhessler:
Thanks for merging! I do not currently have a commercial EPUB example at hand.

For what it is worth, in my own non-commercial scripted EPUB workflow, I use a packaged `.json` file as auxiliary runtime data and load it from a `.js` file via `fetch()`. This passes EPUBCheck 3.3 without complaints in my case.

I realize this is not the commercial example you asked for, but it may still illustrate the kind of EPUB3 scripting workflow where editable packaged JSON resources are useful.

PR #868, opened 2026-05-30
Validate custom XML element names in MetaEditor

BeckyDTP:
Limiting the character set in the name of a custom element. It is better to be cautious than to risk damaging the OPF file, even accidentally.

### DIFF
--- a/src/Dialogs/MetaEditor.cpp
+++ b/src/Dialogs/MetaEditor.cpp
@@ -44,6 +44,7 @@
 #include "Misc/Language.h"
 #include "Misc/MarcRelators.h"
 #include "Misc/SettingsStore.h"
+#include "Misc/Utility.h"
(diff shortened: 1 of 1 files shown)


kevinhendricks:
Will you please add a ':' since xml names may include xml namespace prefixes that end in a colon before the actual name.

Just thought of that case!

Thanks. 

Maybe something like:
```
(?!xml|XML|Xml)[a-zA-Z][a-zA-Z0-9._]*:[a-zA-Z][a-zA-Z0-9._]*|[a-zA-Z]*[a-zA-Z0-9._-]*
```

BeckyDTP:
Hold on. I need to run some tests locally, and then I'll update the commit.

kevinhendricks:
Yes, it is a bit of a nasty regex!  Especially because of the name start char rule!
And please do not forget the missing ; after the hasMatch()

PR #873, opened 2026-07-07
Fix for Preview align after generate HTML TOC

BeckyDTP:


kevinhendricks:
Unfortunately, the bug is that the created file is missing its xml header.  And that bug has existed since forever!

So instead of injecting a newline, add a proper xml header line similar to the following before the doctype.
```
"<?xml version=\"1.0\" encoding=\"utf-8\"?>\n"
```

So please modify your PR and I will merge it.  Thanks.

BeckyDTP:
`m_Writer->writeStartDocument();` is already adding a header, because in both versions (EPUB2 and EPUB3) it adds the following: `<?xml version="1.0"?>`

The missing `encoding="utf-8"` attribute is not a real problem.

I just built a version with that exact change, and everything is working fine.

kevinhendricks:
If an xml header already exists there should be no need to add any newlines.  The bug must be someplace else.

kevinhendricks:
Thank you!

PR #879, opened 2026-08-18
Prevent EPUB import paths from escaping the extraction root

RRRRUDDDD:
## Security issue

EPUB-controlled local paths from `META-INF/container.xml`, package metadata, and encryption metadata were decoded and joined to the extraction directory without proving that the normalized filesystem target remained inside that directory. Manifest MIME detection and later resource loading could read or copy the resulting path before a filesystem boundary check. A crafted EPUB could therefore use relative traversal, absolute paths, Windows drive paths, UNC paths, mixed separators, or encoded path components to make the importer access files outside its extraction root.

## What changed

- Add one shared `Utility::ResolveEPUBPath()` implementation for every EPUB-local filesystem reference.
- URL/XML-decode and normalize references before validation.
- Reject absolute paths, Windows drive-relative and drive-absolute paths, UNC paths, URI schemes, backslashes, and paths that escape after relative-segment normalization.
- Use a separator-aware root boundary check instead of a raw string prefix.
- Canonicalize existing targets; for missing final targets, canonicalize the nearest existing parent. Empty canonical paths and broken symbolic links are rejected.
- Apply the shared validation to container rootfiles, manifest items, metadata links, navigation documents, NCX files, encryption references, MIME sniffing, OPF/NCX rereads, extracted alternate filenames, fonts, and final resource loading.
- Let EPUB imports pass an allowed extraction root to `FolderKeeper::AddContentFileToFolder()` while preserving the existing API behavior for other callers.
- Revalidate EPUB sources before and after existence checks, before `Resource` creation, and immediately before `QFile::copy()`.
- Return worker-thread validation failures to the importing thread as a diagnostic `EPUBLoadParseError` instead of allowing an unhandled worker exception.

## Compatibility

Valid EPUB paths using forward slashes, nested subdirectories, URL-encoded filenames, and relative segments that remain inside the extraction root continue to resolve normally.

### DIFF
--- a/src/BookManipulation/FolderKeeper.cpp
+++ b/src/BookManipulation/FolderKeeper.cpp
@@ -132,14 +132,51 @@ Resource *FolderKeeper::AddContentFileToFolder(const QString &fullfilepath,
                                                bool update_opf, 
                                                const QString &mimetype, 
                                                const QString &bookpath,
-                                               const QString &folderpath)
+                                               const QString &folderpath,
+                                               const QString &allowed_root)
 {
-    if (!QFileInfo(fullfilepath).exists()) {
-        throw(FileDoesNotExist(fullfilepath.toStdString()));
+    QString source_fullpath = fullfilepath;
+    auto validate_epub_source = [&]() {
+        if (allowed_root.isEmpty()) {
+            return;
+        }
+
+        const QString canonical_root = QFileInfo(allowed_root).canonicalFilePath();
+        const QString source_reference = QDir(canonical_root).relativeFilePath(source_fullpath);
+        QString validated_path;
+        QString error_message;
+        if (!Utility::ResolveEPUBPath(canonical_root,
+                                      canonical_root,
+                                      source_reference,
+                                      &validated_path,
+                                      &error_message,
+                                      false)) {
+            throw (EPUBLoadParseError(QString(QObject::tr("Invalid EPUB source path '%1': %2"))
+                                      .arg(source_fullpath, error_message).toStdString()));
+        }
+        source_fullpath = validated_path;
+    };
+
+    // Validate before and after the existence check so that callers cannot
+    // use a source that was resolved outside the extracted EPUB root.
+    validate_epub_source();
+    if (!QFileInfo(source_fullpath).exists()) {
+        if (!allowed_root.isEmpty()) {
+            throw (EPUBLoadParseError(QString(QObject::tr("EPUB resource does not exist inside the extracted root: %1"))
+                                      .arg(source_fullpath).toStdString()));
+        }
+        throw(FileDoesNotExist(source_fullpath.toStdString()));
+    }
+    if (!allowed_root.isEmpty()) {
+        validate_epub_source();
+        if (!QFileInfo(source_fullpath).exists()) {
+            throw (EPUBLoadParseError(QString(QObject::tr("EPUB resource disappeared before loading: %1"))
+                                      .arg(source_fullpath).toStdString()));
+        }
     }
 
     // initialize base file information
-    QString norm_file_path = fullfilepath;
+    QString norm_file_path = source_fullpath;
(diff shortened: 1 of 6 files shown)


kevinhendricks:
Sigil's code already in Utility.cpp and elsewhere unpacks the epub safely and already properly handles checking and preventing unpacking any file outside its target directory which Sigil creates inside its own temporary folder inside the Sigil Preferences folder.

And Sigil already has a URL interceptor that prevent links from outside the book root or Sigil Preferences folder from being loaded.  By default Sigil comes with Javascript and Remote network  access for resources off.

So I am closing this.  I will reopen it once you can supply me with an epub that breaks this containment in some way.  

Relying on others for security patches is really not smart in this day and age without a real testcase proving the existing system is broken and not just misunderstood.

kevinhendricks:
Starting here in ImportEPub.cpp and (duplicated in Utility.cpp)  you will see this code:

https://github.com/Sigil-Ebook/Sigil/blob/master/src/Importers/ImportEPUB.cpp#L477

That literally strips out all upward ../ and  back slash file paths  and *prepends* the absolute path to the target folder *forcing* all files to be unpacked into our own target folder. 

And our URL interceptor already forces all links into our Sigil Preferences folder either into the root extracted path or its enclosing Sigil Preferences folder.

These security change was made long long ago to prevent exactly what you are describing.

Again, please post a evil epub example that gets around this and we will decide on the best fix.  We simply do not accept security based changes from untrusted third parties who do not know our security model nor the code involved.

kevinhendricks:
And as far as I can tell nothing inside META-INF must be touched according to the epub3 spec (just unpacked) except for the container.xml and font encryption.xml and passing xml links through a similar upward and backslash path filter would easily work there as well.   This approach should also work for opf manifest file paths but those files must be found at their paths in order to be unpacked initially as well so that may not be needed.

And no matter what, since Sigil is a local epub file editor, since unpacking is safe, those links could only pull in files the user already has read access to!   These would appear as garbage  in Sigil's epub editor, and would not pass epubchck, not be parseable, etc.  So the user is shown a file he/she can already read.  Without javascript and without remote access, what did this actually accomplish?

kevinhendricks:
For the record:

Tested an epub with an opf manifest that points above the book root path and upon load, this is already detected and the bad file is dropped from the opf and the epub.

Warning: Files exist in the epub that are not listed in the manifest, they will be ignored.
OEBPS/Text/Section0001.xhtml

kevinhendricks:
For the record:

Tested an epub with a META-INF container.xml with a upwards relative path outside the epub root for the full path to the content.opf and received this error on trying to load it in Sigil:

"Cannot Load EPUB: /Users/kebhend/Desktop/bad.epub

No appropriate OPF file found

So neither of the proposed mechansims to somehow access files outside the epub container work with current Sigil.

**And the key point here is:** 

**Even if they did, they could only pull the files the user already had read access to into the epub itself, which would be easily visually detected.  Javascipt is off by default as is all remote access.  So you showed a file to a Sigil user that they already had read access to that file, just like any other local file editor.**

So unless I see a specially crafted epub that actually does damage to the Sigil user in some way (ie. shows them a file that they do not already have read access to)  or somehow ships that file out to some other person without using javascript or network access as they are off by default) then how exactly is this report relevant when the "errors" it is meant to "fix" are already properly handled.

Please focus your "security" attention on software that actually interacts with the network or runs on a server where security actually matters and not on a  single user epub file editor like Sigil.

This is the second "issue" from a supposed security researcher that has been posted to Sigil by untrusted developers.  Hopefully it will be the last.

Please note, we do not accept direct pull requests from untrusted developers that are not reviewed in depth by myself.  And certainly not those related to security!

If you truly feel there is an security issue in Sigil that needs to be addressed, create an issue and attach a working exploit epub with an exploit that actually hurts the user in some way and not actually pulls in a file they could have added themselves since they must have read access to it and it still somehow passes epubcheck!

PR #883, opened 2026-08-29
docs: fix typo wich -> which

vaibhav8a:
Corrects a single misspelling of `wich` in `3rdparty/hunspell/tests/suggestiontest/List_of_common_misspellings.txt`.

Documentation only, no behaviour change.

### DIFF
--- a/3rdparty/hunspell/tests/suggestiontest/List_of_common_misspellings.txt
+++ b/3rdparty/hunspell/tests/suggestiontest/List_of_common_misspellings.txt
@@ -3962,7 +3962,7 @@ wholey	wholly
 wholy	wholly, holy
 whta	what
 whther	whether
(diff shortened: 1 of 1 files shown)


vaibhav8a:
Closing this — my mistake. The only match was inside `3rdparty/hunspell/tests/suggestiontest/List_of_common_misspellings.txt`, which is a vendored Hunspell test fixture where `wich` is deliberately the misspelled input. Sorry for the noise.

kevinhendricks:
Two things:

- you should not post issues from 3rdparty code to Sigil, instead post them directly to the original project

-Sigil does not accept pull requests or issues that fix spelling in our code that is not in a string that the user actually sees ( and only in English, not translations).   And of course no fixes for spelling in code comments are ever accepted.

Issue #882, opened 2026-08-28
[Bug]: Help → About section contains outdated Transifex link


cndxflwb:
## Problem
The "Translators" link in Sigil's Help → About dialog points to an outdated Transifex URL that no longer works.

## Steps to Reproduce
1. Open Sigil.
2. Go to `Help` → `About`.
3. Click on the "Translators" link (or note the URL displayed).

## Actual Result
The link `https://www.transifex.com/projects/p/sigil/` returns a **404 Not Found** error.

## Expected Result
The link should direct users to the active translation project page.

## Suggested Fix
Update the URL to:  
`https://explore.transifex.com/zdpo/sigil/`

This is the current active translation project page for Sigil on Transifex.

## Additional Context
- The old Transifex projects URL structure (`/projects/p/`) has been deprecated.
- The new URL follows Transifex's current Explore platform structure.
- The new page confirms Sigil's translation project with 42 languages actively maintained.

## Affected Versions
All versions that still use the old Transifex URL format (likely all current releases).

## Possible Code Locations
The URL is likely defined in the About dialog source code or in a translation-related configuration file.

### Platform (OS)

Windows (Default)

### OS Version / Specifics

Windows 10

### What version of Sigil are you using?

2.8.1

### Any backtraces or crash reports

```shell

```

dougmassay:
The Transifex url has been update to the correct one. Thank you for your bug report.

I took me two commits because I'm an idiot sometimes!

e3cae2a
478041b

PR #887, opened 2026-08-30
Two additional links to the Transifex website

BeckyDTP:
In addition to the "About" window, there are two other mentions of Transifex, so this is a unification PR.

Additional fix for #882


dougmassay:
Good call. Thank!

PR #888, opened 2026-09-10
Crop Feature Improvements and New Refresh Function

byword77:
1. Images are now centered in the viewer screen.
2. Enhanced the crop feature for better usability.
- Press CTRL + Left Mouse Click to set the coordinates -> Click the Crop button.
3. Added a refresh function.
- You can now update the currently opened image file in the viewer after modifying it with an external editor or overwriting it by adding the same image file.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -1,16 +1,20 @@
-Sigil 2.8.x
+Sigil 2.8.1
    Bug Fixes
-     - fix bug in AddExisting that could sometimes add duplicate image entries and skip some xhtml imports
+     - fix multiple bugs in AddExisting that could sometimes add duplicate image entries and skip some xhtml imports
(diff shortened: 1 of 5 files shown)


kevinhendricks:
Sigil's dev language is English so you need to fix all commit comments to be English.

Please do not add in any additions not related to the improved crop functionality.

kevinhendricks:
Please focus this on the crop functionality, not centering, not refresh, just the crop as requested.

Opening the same Image file in two different editors at the exact same time is just a really bad idea.  I will convert a "open with" to close any associated Image tab so no refresh is needed.

You should not need 19 commits.

Perhaps a diff would be better than a PR here after all.

byword77:
I deleted this repository because the code got messy from various tests. I will create a fresh repository based on the current version, apply only the changes you mentioned, and push it again.

PR #889, opened 2026-09-11
Crop Method Improvements

byword77:
1. Set Crop Area: Hold Ctrl + Left-click and drag to define the crop coordinates.
2. Adjust Coordinates: Click and drag any of the four corners of the selected area to modify the coordinates.
3. Cancel Crop Area: Hold Ctrl and click anywhere inside the image area to cancel the selected coordinates.
4. Execute Crop: Click the Crop button while the coordinates are set to crop the image to the specified area.

Could you take a look at the code? I'm not a professional developer, so feel free to reject it if you spot any issues.

### DIFF
--- a/src/Widgets/AdjustImage.cpp
+++ b/src/Widgets/AdjustImage.cpp
@@ -33,6 +33,8 @@
 #include <QTransform>
 #include <QDebug>
 #include <QFileInfo>
+#include <QFile>
+#include <QLocale>
 #include <QImageWriter>
 #include <QInputDialog>
 #include <QKeySequence>
@@ -154,6 +156,8 @@ AdjustImage::AdjustImage(const QString filepath, const QString& mediatype,  QWid
     setWindowTitle(tr("Adjust Image"));
     if (!filepath.isEmpty()) {
         m_fileName = filepath;
+        m_ffsize = QFile(m_fileName).size() / 1024.0;
+        m_fsize =  QLocale().toString(m_ffsize, 'f', 2);
         m_image = QImage(m_fileName);
         if (m_image.isNull()) {
              QMessageBox::information(this,
@@ -161,6 +165,7 @@ AdjustImage::AdjustImage(const QString filepath, const QString& mediatype,  QWid
                                       tr("Cannot load %1.").arg(m_fileName));
              return;
         }
+        m_scaleFactor = 1.0;
         setCursor(Qt::ArrowCursor);
         updateActions(true);
         refreshLabel();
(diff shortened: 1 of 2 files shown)


kevinhendricks:
Getting closer.

Unfortunately, your version seems to be removing recent changes to AdjustImage that show file size on the status bar, and removing scale factors which should not be removed.

Please start with *current* Sigil master versions of those files, then only add the parts you need to add without removing existing code that is unrelated to your changes.

kevinhendricks:
Perhaps, it might be easier to diff your updated AdjustImage .ui, .cpp, and .h files against the exact versions from Sigil that you based them on initially.  Then we can see exactly what you want to add without the spurious code removals.

byword77:
I was working on v2.8.1, so I missed the changes in v2.8.5. I've now checked the updates in v2.8.5 and updated everything accordingly.

kevinhendricks:
Instead of basing your patches and changes on our current master, you based them on a previous version, then added a commit to put back things that never should have been removed in the first place.  Furthermore your PR change things like how we use QRubberBand mode, how things are initialized and when -  that did not need to be changed, installs event filters all over the place, add classes for things like corner points that may not be truly needed.

**So I simply can not tell if what you are proposing actually works without breaking things.**

Given, I would have to later support these changes, I am not going to accept this PR without some serious testing and then cleanups to greatly reduce their intrusiveness, especially just for changing how crop points are selected.

So I will leave this PR open.  I will try out the code in my own builds and try to minimize their intrusiveness and remove all the spurious changes.

If at the end I can get something that works, looks supportable, does not break anything, and does not include spurious changes, I will pass an acceptable version  along to you and you can make a better PR which we can accept.  One of the problems with AI generated code is that it does not understand broad design choices made across the project, does not understand supportability, does not understand the importance of minimizing changes, especially spurious ones, does not follow one class one file design, etc.

Thank you.

kevinhendricks:
After staring at the code, I think we can subclass QRubberBand and move most of your Corner handle code into that subclass simplifying the AdjustImage which will use this subclass to replace its use of QRubberBand class.

Hopefully, it will simplify things.  I will try to get something working so that you can use it to create an improved PR.

byword77:
내용을 이해했습니다. 처음으로 git을 이용해 코드를 올려봤는데 배운게 많았습니다. 여기까지가 제 한계인 것 같고, 실력있는 분들이 코드를 개선해 Sigil을 이용하는 사람들에게 도움이 되었으면 좋겠어요.

kevinhendricks:
Okay, I will take it from here and still credit you to the extent I can for your ideas and code.

kevinhendricks:
Okay I have pushed a version of this based on subclassing QRubberBand that actually simplifies AdjustImage.cpp
Hitting Crop now puts you in crop mode, where you can use Enter to crop to your selection or Escape to abort.

If people like this improvement it will appear in the next Release of Sigil.
Thank you for your PR and hard work on this!

